PR #4, opened 2020-02-19
[New] Built basic app file structure

Rizvida1:
- Installed Ant Design
- Created example file structure with components and pages


dynamic11 (approved):
I approved it, but we should sit down and go through it tomorrow so you can give us a game plan

PR #24, opened 2020-02-26
[NEW] Added toggle for change language

Rizvida1:
- Change language is now a toggle


dynamic11:
Please hold this change.
I really like it but we somehow need to make that toggle smaller.
Currently our navbar is too long for small screen sizes (even with our current button).
We can discuss it when we see each other next.

PR #32, opened 2020-03-04
[New] Admin User Table Page

SukhsimranpreetSekhon:
- Allows for admin to see existing users on the app and their details.
- It allows easy access to a user's profile page.
- Dropdown options allow admin to make a user active, inactive or hidden


services/frontend-v3/package.json
```diff
@@ -18,6 +18,7 @@
     "react": "^16.12.0",
     "react-app-rewired": "^2.1.5",
     "react-dom": "^16.12.0",
+    "react-highlight-words": "^0.16.0",
```
dynamic11, line 21: is it possible to avoid importing this? I want to minimize the number of packages we use to avoid future support issues.

dynamic11 (changes requested):
looks good just see if you can avoid the problem i mentioned in the comment. we can talk about it together tomorrow. I will be in the office around 10.

services/frontend-v3/package.json
SukhsimranpreetSekhon: Yeah, we need it for the searching, where it will highlight the letters/numbers that the admin searched for in a column. I think that is the last new technology that I will need for the admin pages.

dynamic11 (approved):

PR #51, opened 2020-03-11
[New] Redirect user to sign up page

dynamic11:
- force a redirect to set up page if a profile is not found
- if the user is already on the profile page then avoid redirect
- clean up and comment file

### DIFF
--- a/services/frontend/src/components/routes/Secured.js
+++ b/services/frontend/src/components/routes/Secured.js
@@ -65,6 +65,8 @@ class Secured extends Component {
   goto = link => history.push(link);
 
   render() {
+    console.log("helloooo");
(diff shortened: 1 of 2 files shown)


services/frontend/src/components/routes/Secured.js
```diff
@@ -65,6 +65,8 @@ class Secured extends Component {
   goto = link => history.push(link);
 
   render() {
+    console.log("helloooo");
+    console.log(this.state);
```
Rizvida1: Should these be here?

Rizvida1 (changes requested):

Rizvida1 (approved):

PR #55, opened 2020-03-12
[Update] Changed SearchBar to function component and added styling to profile

KateAThornley:
- updated searchbar.jsx to function component instead of class
- adjusted ant design default fonts to be larger in app.css
- added styling to cards and tags on profile page


s-glmxmd (approved):
Good job Kat! 😊

KateAThornley:
- updated searchbar.jsx to function component instead of class
- adjusted ant design default fonts to be larger in app.css
- added styling to cards and tags on profile page

PR #62, opened 2020-03-17
[New] Added sider to profile page

Rizvida1:
- There is now a sider anchor on profile page

### DIFF
--- a/services/frontend-v3/src/i18n/en_CA.json
+++ b/services/frontend-v3/src/i18n/en_CA.json
@@ -136,7 +136,7 @@
   "profile.acting.period.start.date": "Acting start date",
   "profile.address": "Address",
   "profile.branch": "Branch",
-  "profile.basic": "Basic Info",
+  "profile.basic": "Employee Information",
   "profile.diploma": "Diploma",
   "profile.projects": "Projects",
   "profile.school": "School",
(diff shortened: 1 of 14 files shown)


services/frontend-v3/src/i18n/en_CA.json
```diff
@@ -136,6 +136,7 @@
   "profile.acting.period.start.date": "Acting start date",
   "profile.address": "Address",
   "profile.branch": "Branch",
+  "profile.basic": "Basic Info",
```
s-glmxmd: Where is this used? If its for the main card I don't think we need a title for the card.

services/frontend-v3/src/i18n/fr_CA.json
```diff
@@ -139,6 +139,7 @@
   "profile.acting.period.start.date": "Date de début intérimaire",
   "profile.address": "Adresse",
   "profile.branch": "Branche",
+  "profile.basic": "Info de base",
```
s-glmxmd: Same comment for english... not sure why where this is used or why we would need one.

services/frontend-v3/src/i18n/en_CA.json
s-glmxmd: Would recommend Personal Information or Employee Information

services/frontend-v3/src/i18n/fr_CA.json
s-glmxmd: Translation would be Information personnel or Information d'employée

services/frontend-v3/src/i18n/en_CA.json
Rizvida1: It's used for the anchor thing I made

s-glmxmd (approved):
Good job Rizvi! 😄

PR #67, opened 2020-03-19
[Fix] Removed Global Styling

KateAThornley:
- removed global styling that was tampering with other parts of app


s-glmxmd (approved):
Impeccable work once again Kat! 👏

PR #69, opened 2020-03-19
[FIX] Private Info variable missing declaration

s-glmxmd:
'Twas an easy fix 😥


KateAThornley (approved):
Sagal's back at it again with the incredible bug fixes🤩

PR #72, opened 2020-03-22
[New] Created talent form

dynamic11:
- added field for competencies
- added field for skills
- added field for skills for mentorship (these skills are limited to the skills selected in the skills field.

### DIFF
--- a/services/backend/server.js
+++ b/services/backend/server.js
@@ -79,9 +79,9 @@ router.get("/getEmployeeInfo/:searchValue", keycloak.protect(), async function(
 });
 
 //User endpoints
-router.get(keycloak.protect(), "/user/", user.getUser);
-router.get(keycloak.protect(), "/user/:id", user.getUserById);
-router.post(keycloak.protect(), "/user/", user.createUser);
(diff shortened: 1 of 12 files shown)


services/backend/server.js
```diff
@@ -79,17 +79,17 @@ router.get("/getEmployeeInfo/:searchValue", keycloak.protect(), async function(
...
+router.get("/user/:id", user.getUserById);
+router.post("/user/", user.createUser);
 
 //Profile endpoints
 router.get("/profile/", keycloak.protect(), profile.getProfile);
 router
   .route("/profile/:id")
-  .get(keycloak.protect(), profile.getPublicProfileById)
-  .post(keycloak.protect(), profile.createProfile)
-  .put(keycloak.protect(), profile.updateProfile);
```
Rizvida1: Be careful. Our app is now live so stuff like this is risky

Rizvida1 (changes requested):

services/backend/server.js
dynamic11: sorry i keep forgetting to put this back :(

PR #79, opened 2020-03-30
[Fix] Added Missing Accessibility Features

KateAThornley:
-added error message on empty home search
-added warning message when no results are found on home page
-added tabbing to home page (minus to toggle the menu down in topNav- still working on that)
-added tabbing to results page including advanced search filters in sidebar
-checked all contrast, labels, and wave tool errors for accessibility

### DIFF
--- a/services/frontend-v3/src/components/searchBar/SearchBarView.jsx
+++ b/services/frontend-v3/src/components/searchBar/SearchBarView.jsx
@@ -1,15 +1,24 @@
 import React, { useState } from "react";
 import { injectIntl } from "react-intl";
 import { Form, Alert } from "antd";
-import { Icon as LegacyIcon } from "@ant-design/compatible";
+import { DownOutlined, UpOutlined } from "@ant-design/icons";
+//import { Icon as LegacyIcon } from "@ant-design/compatible";
 import "@ant-design/compatible/assets/index.css";
 import { Row, Col, Button, Card } from "antd";
 import logo from "../sideNav/logo_v2.svg";
 
 function SearchBarView(props) {
-  const [expand, setExpand] = useState(false);
+  const [setDoNothing] = useState(true);
   const [form] = Form.useForm();
-  const { getFields, getBasicField, handleSearch, toggle, data, empty } = props;
+  const {
+    getFields,
+    getBasicField,
+    handleSearch,
+    expand,
+    toggle,
+    data,
+    empty
+  } = props;
 
   const searchLabel = props.intl.formatMessage({
     id: "button.search",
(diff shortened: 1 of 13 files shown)


services/frontend-v3/src/components/searchBar/SearchBarView.jsx
```diff
@@ -1,15 +1,24 @@
...
+import { Form, Alert } from "antd";
+import { DownOutlined, UpOutlined } from "@ant-design/icons";
+//import { Icon as LegacyIcon } from "@ant-design/compatible";
 import "@ant-design/compatible/assets/index.css";
 import { Row, Col, Button, Card } from "antd";
 import logo from "../sideNav/logo_v2.svg";
 
 function SearchBarView(props) {
-  const [expand, setExpand] = useState(false);
+  const [setDoNothing] = useState(true);
```
Rizvida1: This should not work...
The useState returns an array that looks like [state, setState]. You are setting setDoNothing as the state and not the setState. 
This would work with objects but not arrays.

Rizvida1 (changes requested):

services/frontend-v3/src/components/searchBar/SearchBarView.jsx
```diff
@@ -20,17 +29,29 @@ function SearchBarView(props) {
...
+        <Alert
+          message={props.intl.formatMessage({
+            id: "alert.empty.search",
+            defaultMessage: "Please input a value into the search bar below"
+          })}
+          type="error"
+          style={styles.alert}
+        />
+      ) : (
+        setDoNothing[true]
```
Rizvida1, line 53: That's not how it should work and you probably do get the warning. if you want, you can do temp = useState... and then do setDoNothing=temp[1]

Rizvida1 (changes requested):

services/frontend-v3/src/components/searchBar/SearchBarView.jsx
KateAThornley: wait why should my setDoNothing not be allowed to be set to true? If i create a temp state using useState then am I not going to need setTemp?? I think im missing something here...

Rizvida1: You aren't setting here. First of all, you are using square brackets to call a function. Secondly, setDoNothing was not a function before. When you use useState, the names of the variables don't matter, the order does since it is an array, not an object. And the method I suggested above is to remove the error associated with what you're doing. It is better now since you added "doNothing" to the useState but you'll get the error for not using in and you still aren't calling "setDoNothing" and a function.

Rizvida1 (changes requested):

dynamic11:
branch has fallen behind (stale)

PR #80, opened 2020-04-02
[New] Created new landing page

KateAThornley:
- Created LandingPage.jsx
- Styled with new banner (let me know if we'd like it replaced)
- Added logo and language button to top nav of landing page
**Note: waiting on info from clients on what content we'd like to fill the landing page with, when that comes in I will add the proper cards to the page


dynamic11:
It looks sweet! I just have a few questions:

How temporary is this home page? The reason I am asking is that there are two problems with it that i found but I'm not sure if its worth the time to fix it. here are my concerns:

1) its not responsive and im worried that even on our surfaces it will look unresponsive
2) i like the style but i think some people will have issues with how dark it is. We can mock up some other styles but i don't want to waste time if its all going to be changed in the future.

Rizvida1 (approved):

PR #81, opened 2020-04-02
[new] updated antd to 4.1

dynamic11:
- update antd to 4.1
- update other related packages


Mousto097 (approved):
It looks good!

I checked out your branch, cleaned up all containers, I built the whole app and everthing was working perfectly.

PR #82, opened 2020-04-02
[New] Display only public profile cards when visiting other users profile

Mousto097:
- Added ability to check wether current user looking at his own profile or others profile
- Request backend data (private or public data) depending on what profile current user is trying to access
- Change some function name to more descriptive one
- Handle displaying two card on the same row

### DIFF
--- a/services/frontend-v3/src/components/competenciesCard/Competencies.jsx
+++ b/services/frontend-v3/src/components/competenciesCard/Competencies.jsx
@@ -11,10 +11,12 @@ class Competencies extends Component {
     let competencies = {};
     let key = 0;
 
-    data.competencies.forEach(element => {
-      competencies[key] = element.description[locale];
-      key++;
-    });
+    if (data.competencies) {
+      data.competencies.forEach(element => {
+        competencies[key] = element.description[locale];
+        key++;
+      });
+    }
 
     return competencies;
   }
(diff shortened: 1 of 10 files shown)


services/frontend-v3/src/components/competenciesCard/Competencies.jsx
```diff
@@ -11,10 +11,12 @@ class Competencies extends Component {
     let competencies = {};
     let key = 0;
 
-    data.competencies.forEach(element => {
-      competencies[key] = element.description[locale];
-      key++;
-    });
+    if (data.competencies) {
```
dynamic11, line 14: everything looks good just out of curiosity why did you put in this protection? why was it not necessary before. I dont think its wrong (i actually like it) i just want to know what made you do it.

dynamic11 (approved):
Please answer my question(i left one comment)
It looks good though!

services/frontend-v3/src/components/competenciesCard/Competencies.jsx
Mousto097: When a user hides his competencies, data.competencies is null (getPublicProfileById sends null for hidden data). And when data.competencies is "null", forEach throws an error.

The error was "Uncaught TypeError: Cannot read property ‘forEach’ of null"

PR #83, opened 2020-04-03
[FIX] Conversion to functional components

s-glmxmd:
Converted components to functional instead of class.
Went through code to ensure we were respecting MVC model for the files that needed to be converted.


services/frontend-v3/src/components/careerInterests/CareerInterests.jsx
```diff
@@ -1,14 +1,56 @@
...
+      data={props.data}
+      locale={localStorage.getItem("lang")}
+      info={getCareerInterestsInfo(props)}
+      relocationLocationsInfo={getRelocationLocationsInfo(props)}
+    />
+  );
 }
 
-export default CareerInterests;
+export default injectIntl(CareerInterests);
```
Rizvida1: Are you adding injectIntl when you aren't using props.intl?

Rizvida1 (changes requested):

services/frontend-v3/src/components/careerInterests/CareerInterests.jsx
s-glmxmd: Yes.. unfortunately I was. 😞  Fixed it tho!

Rizvida1 (approved):

PR #93, opened 2020-04-07
[Fix] Fixed issue related to visible cards on the profile page

Mousto097:
- Problem: whether you are on your own profile or not, we are using current user data to display the cards.
- Solution: fixed it by using the urlID (which changes dynamically depending on whose profile is being look at) instead of always using the same locolstorageId.


s-glmxmd (approved):
Looks good! 😄 👍

PR #100, opened 2020-04-13
[New] Added Coloured Initials to Results Cards

KateAThornley:
- Added avatarColor and nameInitials to getAllProfiles.js in backend
- Added avatars to results cards in ResultCardView

### DIFF
--- a/services/frontend-v3/src/components/basicInfo/BasicInfoView.jsx
+++ b/services/frontend-v3/src/components/basicInfo/BasicInfoView.jsx
@@ -10,20 +10,20 @@ function BasicInfoView(props) {
   /* Component Styles */
   const styles = {
     profileHeaderRow: {
-      margin: "25px 0"
+      margin: "25px 0",
     },
     card: {
       borderTopColor: "#007471",
-      borderTopWidth: "5px"
+      borderTopWidth: "5px",
     },
     avatar: {
       backgroundColor: "#fff",
       color: "#007471",
-      marginRight: "-10px"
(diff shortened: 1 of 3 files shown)


services/frontend-v3/src/components/basicInfo/BasicInfoView.jsx
```diff
@@ -10,20 +10,20 @@ function BasicInfoView(props) {
   /* Component Styles */
   const styles = {
     profileHeaderRow: {
-      margin: "25px 0",
```
dynamic11: I think we decided to leave these commas in. As trailing commas are recommended by some people to make diffs easier. We should discuss this with Risvi further.

dynamic11 (changes requested):
We should discuss the comment as a team to come up with a solution so that it can become a standard.

services/frontend-v3/src/components/basicInfo/BasicInfoView.jsx
Rizvida1: I agree. Maybe your version of Prettier or VS Code is not up to date, Kat.

dynamic11: maybe we should add a prettier config file to resolve this issue in the future?

KateAThornley: I fixed this- my prettier was out of date sorry guys!

dynamic11 (approved):

PR #110, opened 2020-04-19
[Fix] Add french form options

dynamic11:
- added translation for profile form dropdowns
- moved antd code out of layout controllers for the forms
- restyled form save button to work on smaller screens
- cleaned up warnings


services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
```diff
@@ -483,18 +499,16 @@ const EmploymentDataFormView = (props) => {
                   <Select
                     showSearch
                     optionFilterProp="children"
-                    placeholder="choose classification"
+                    placeholder={<FormattedMessage id="admin.select" />}
```
Rizvida1: Should you be using a newly created id? If someone decides to change the one for admin, it might not work for this page anymore.

Rizvida1 (changes requested):

services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
dynamic11: will change :)

PR #111, opened 2020-04-20
[New] Added competencies table/page

SukhsimranpreetSekhon:
- Created competencies table feature for Admin side
- Contains competencyTable component (controller & view)
- Allows to add, edit and delete competencies
- Can also search through pre-existing competencies


services/frontend-v3/src/components/competencyTable/CompetencyTable.jsx
```diff
@@ -0,0 +1,173 @@
...
+    let allCompetencies = _.sortBy(data, description);
+
+    for (let i = 0; i < allCompetencies.length; i++) {
+      allCompetencies[i].key = allCompetencies[i].id;
+    }
+
+    return allCompetencies;
+  };
+
+  useEffect(() => {
```
Rizvida1: Generally its safer for hooks to be closer to the top of the file

Rizvida1: It's because they need to be created every time the function runs and if anything stops the function before a hook, it throws an error

SukhsimranpreetSekhon: Fixed.

Rizvida1 (approved):

PR #120, opened 2020-04-22
[New] Created Admin Diploma table/page & Fixed delete confirmation language (English)

SukhsimranpreetSekhon:
- Added Diploma table/page to Admin side
- Able to add, edit and delete diploma information
- Can search through diploma information
- English/French translation
- Added access from the sidebar to diploma page
- Fixed delete confirmation language (diploma & competency table pages)


services/frontend-v3/src/components/diplomaTable/DiplomaTable.jsx
```diff
@@ -0,0 +1,172 @@
...
+const backendAddress = config.backendAddress;
+
+function DiplomaTable(props) {
+  const [data, setData] = useState([]);
+  const [loading, setLoading] = useState(true);
+  const [reset, setReset] = useState(false);
+  const [searchText, setSearchText] = useState("");
+  const [searchedColumn, setSearchedColumn] = useState("");
+  const [selectedRowKeys, setSelectedRowKeys] = useState([]);
+  const [size] = useState("large");
```
dynamic11: if you are never going to set the sate then why use a state. you can just make variable. States should only be used for variables that will be updated (and thus would need a rerender)

services/frontend-v3/src/components/diplomaTable/DiplomaTable.jsx
```diff
@@ -0,0 +1,172 @@
...
+
+  const handleSearch = (selectedKeys, confirm, dataIndex) => {
+    confirm();
+    setSearchText(selectedKeys[0]);
+    setSearchedColumn(dataIndex);
+  };
+
+  const handleReset = (clearFilters) => {
+    clearFilters();
+    setSearchText("");
```
dynamic11, line 74: you can just set the field value rather then use a state

services/frontend-v3/src/components/diplomaTable/DiplomaTable.jsx
```diff
@@ -0,0 +1,172 @@
...
+  };
+
+  const rowSelection = {
+    onChange: (selectedRowKeys) => {
+      onSelectChange(selectedRowKeys);
+    },
+  };
+
+  const onSelectChange = (selectedRowKeys) => {
+    // console.log("selectedRowKeys changed: ", selectedRowKeys);
```
dynamic11: please remove

services/frontend-v3/src/components/diplomaTable/DiplomaTable.jsx
```diff
@@ -0,0 +1,172 @@
...
+    setSelectedRowKeys(selectedRowKeys);
+  };
+
+  const convertToViewableInformation = () => {
```
dynamic11, line 133: please comment the purpose of each function above it

services/frontend-v3/src/components/diplomaTable/DiplomaTable.jsx
SukhsimranpreetSekhon: Fixed.

services/frontend-v3/src/components/diplomaTable/DiplomaTable.jsx
SukhsimranpreetSekhon: I leave that console.log in every table, just in case one day, the select and delete does not work out, so we can see why.

services/frontend-v3/src/components/diplomaTable/DiplomaTable.jsx
SukhsimranpreetSekhon: Yeah, need to comment on some other things from Admin side too.

services/frontend-v3/src/components/diplomaTable/DiplomaTable.jsx
SukhsimranpreetSekhon: Resolved.

services/frontend-v3/src/components/diplomaTable/DiplomaTable.jsx
SukhsimranpreetSekhon: I am thinking to do all the commenting on one branch and then PR it?

dynamic11: i would do as much commenting as you can related to this PR. in case your next PR gets reverted for some reason. Its just best practice.

dynamic11 (approved):

PR #139, opened 2020-04-25
[Fix] frontend docker container crash

### LINKED ISSUE #132: Docker container for frontend failing on Run
- Docker fails to run the `upskill_web-ant` container (frontend)
- the container crashes Run
- this issue has been observed since pull request #118 
- we suspect it is due to update of react-scripts from `3.4.0` to `3.4.1`

*** Disclaimer: If you run the container in interactive mode, the web app will work. Unsure at the moment for this case. ****

dynamic11:
- fixes #132 caused by upgrading yarn packages (PR: #118 )
- the container image would **Build** but would fail right away on **Run**
- rollback react-scripts 3.4.0

### DIFF
--- a/services/frontend-v3/package.json
+++ b/services/frontend-v3/package.json
@@ -23,7 +23,7 @@
     "react-highlight-words": "^0.16.0",
     "react-intl": "^3.12.0",
     "react-router-dom": "^5.1.2",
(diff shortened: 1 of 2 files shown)


Mousto097 (approved):
Looks gooood!

PR #154, opened 2020-05-05
[Fix] Numbering issue in edit profile

### LINKED ISSUE #148: incorrect numbering in edit from headings
The numbering for the form heading jumps from 1 to 3.
_**NOTE**_: probably a typo in the profile forms layout

### LINKED ISSUE #159: Forms with short height have large empty space at bottom of form card
**Describe the bug**
If a profile form is short the card that holds the form does not shrink to the height of the form so it leaves large blank space at the bottom of the card

**To Reproduce**
1. Go to edit profile
2. on side bar select language proficiency
3. see big gap at bottom of card

**Expected behavior**
the height of the card should match the height of the form to avoid empty space

**Screenshots**
image

BenJeau:
- remove "step" numbering in the header when editing a profile (closes #148)
- make edit profile form card match height of content (closes #159)
- remove redundant <div>

### DIFF
--- a/services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
+++ b/services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
@@ -39,11 +39,16 @@ const EmploymentDataFormView = (props) => {
 
   /* Component Styles */
   const styles = {
+    skeleton: {
+      minHeight: "400px",
+      maxWidth: "900px",
+      background: "#fff",
+      padding: "30px 30px",
+    },
     content: {
       textAlign: "left",
       width: "100%",
       maxWidth: "900px",
-      minHeight: "400px",
       background: "#fff",
       padding: "30px 30px",
     },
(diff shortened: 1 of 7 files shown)


services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
```diff
@@ -446,7 +446,7 @@ const EmploymentDataFormView = (props) => {
     return (
       <div style={styles.content}>
         <Title level={2} style={styles.formTitle}>
-          3. <FormattedMessage id="setup.employment" />
```
dynamic11, line 449: Changing the numbering here will break it for create profile since it is a shared form. Maybe we should remove numbering or pass it in from the layout page?

BenJeau: I just tested (nuked the db) and when going through the create profile flow, the numbering was still okk surprisingly

PR #164, opened 2020-05-06
[Fix] Advanced Search for Skills

### LINKED ISSUE #153: Advanced searching with skills doesnt return results
**Describe the bug**
When searching for skills via the advanced search on the home page or in the sidebar for the results page, no results are returned even if there are profiles matching the skill search.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to homepage of I-Talent
2. Search for a skill that is on a profile in the system in the advanced search (ex: java) 
3. Sends user to results page with no results

**Expected behaviour**
Should be returning cards with matching users to search criteria, including the tags of the relevant skills on the profile cards

KateAThornley:
- Fixed typo on privateinfo skills which blocked skills from being returned to search

### DIFF
--- a/services/backend/API/search/util/skillSearch.js
+++ b/services/backend/API/search/util/skillSearch.js
@@ -1,5 +1,4 @@
 const skillSearch = async (profiles, skillarray) => {
-  console.log("AHHHHHHHHHHH");
   let skillProf = [];
   profiles.forEach((profile) => {
     if (!profile.skills) return;
--- a/services/backend/API/search/util/skillSearch.js
+++ b/services/backend/API/search/util/skillSearch.js
@@ -1,10 +1,11 @@
 const skillSearch = async (profiles, skillarray) => {
+  console.log("AHHHHHHHHHHH");
   let skillProf = [];
-  profiles.forEach(profile => {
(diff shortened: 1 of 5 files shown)


services/frontend-v3/src/components/resultsCard/ResultsCard.jsx
```diff
@@ -25,14 +25,12 @@ function ResultsCard(props) {
       let queryString = urlSections[1];
       axios
         .get(backendAddress + "api/search/fuzzySearch?" + queryString)
-        .then(result => setResults(result.data));
-      // gatherResults(queryString);
+        .then((result) => setResults(result.data));
+      //gatherResults(queryString);
```
dynamic11: should we remove comment?

services/backend/API/search/util/skillSearch.js
```diff
@@ -1,10 +1,11 @@
 const skillSearch = async (profiles, skillarray) => {
+  console.log("AHHHHHHHHHHH");
```
dynamic11: remove

dynamic11 (changes requested):

dynamic11 (approved):

PR #169, opened 2020-05-06
Add expandable description for job experience items

### LINKED ISSUE #134: Add "Expand for Description" on Experience cards (User Profile)
- See Kat for mockup
- Add Antd expand component under each experience entry

TrevorBivi:
-changed resultsCard's left orientated divider changing theme of all left orientated dividers in the app
- added expandable job description to profile
- closes #134 
note: I added a class name for resultsCard dividers to use and altered the CSS changes in app.css to only affect left orientated dividers that also have that class name instead of all left orientated dividers

description dividerless real

### DIFF
--- a/services/frontend-v3/src/components/experience/experienceItem/ExperienceItemView.jsx
+++ b/services/frontend-v3/src/components/experience/experienceItem/ExperienceItemView.jsx
@@ -10,7 +10,8 @@ function ExperienceItem(props){
         },
         experienceDescription: {
             color: "rgba(0, 0, 0, 0.85)",
-            maxWidth:"525px"
+            maxWidth:"525px",
+            paddingTop:"6px"
         },
         experienceDescriptionToggleTag: {
             color: "rgba(0, 0, 0, 0.85)",
@@ -19,9 +20,6 @@ function ExperienceItem(props){
         expandDescriptionToggleTagText:{
             paddingLeft:"5px"
         },
-        experienceDescriptionDivider:{
-            fontSize:"14px"
-        }
       };
     const {expand, item, toggleExpand} = props;
     
@@ -29,12 +27,6 @@ function ExperienceItem(props){
             if (expand) {
                 return (
                     <Row>
-                        <Divider plain style={styles.experienceDescriptionDivider} className="experienceDescriptionDivider" orientation="left">
-                        {props.intl.formatMessage({
-                                id: "profile.career.content.title",
-                                    defaultMessage: "Job Description",
-                                })}
-                        </Divider>
                         <p style={styles.experienceDescription}>{item.description}</p>
                     </Row>);
             } else {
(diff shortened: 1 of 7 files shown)


TrevorBivi:
~~Here is an image of what I implemented. A picture of the mockup can be found in the issue https://puu.sh/FHa5C.png~~ Just realised you can upload images directly to github and that looks way nicer so I added one to the description

services/frontend-v3/src/components/experience/experienceItem/ExperienceItemView.jsx
```diff
@@ -0,0 +1,85 @@
...
+import { ContainerOutlined, DownOutlined, UpOutlined } from "@ant-design/icons";
+
+function ExperienceItem(props){
+    const styles = {
+        avatar: {
+          backgroundColor: "#007471",
+        },
+        experienceDescription: {
+            color: "rgba(0, 0, 0, 0.85)",
+            maxWidth:"525px"
```
BenJeau: is there a reason to make the text container width have a limit? I feel like it looks weird not having it fill the space. Aside from that, everything seems good :100:

TrevorBivi: @KateAThornley made a mockup that seemed to have the max width limited to around 525px. Maybe she can respond and say if there was reason for that or not. The widest version of it shown in the above image is on a 1440p monitor, bigger than most govt employees would be using, so that makes it look a fair bit weirder

BenJeau: ah okok, I just tried it locally and was wondering why it was not the full width

dynamic11:
why do we need heading for job description? I feel like it is self-explanatory no? Im also not of the big fan of the heading with the line because it make it look busy. I think the implementation looks good but im not a big fan of the line thing that antd does

services/frontend-v3/src/components/experience/experienceItem/ExperienceItemView.jsx
KateAThornley: O noo in my mockup lorem ipsum had an enter space in it so it was a newline not a max width shoot my bad I shiouldve specified that

KateAThornley:
@dynamic11 

> why do we need heading for job description? I feel like it is self-explanatory no? Im also not of the big fan of the heading with the line because it make it look busy. I think the implementation looks good but im not a big fan of the line thing that antd does

Thats totally fair! I honestly like the dividers but its true that its pretty self-explanitory

dynamic11:
@KateAThornley honestly it just comes down to personal preference. So I'm not sure. @BenJeau any opinions?

TrevorBivi:
used ms paint to make a version without the header to see what it would look like better
description dividerless

BenJeau:
I usually like more of a minimalist look, so I'll have to go with @dynamic11

KateAThornley:
> used ms paint to make a version without the header to see what it would look like better
> ![description dividerless](https://user-images.githubusercontent.com/30152653/81332522-339f3f00-9071-11ea-8831-b3e15005166d.png

> I usually like more of a minimalist look, so I'll have to go with @dynamic11

Yeah i think it looks good w/o the divider!

dynamic11 (approved):

PR #196, opened 2020-05-07
[Feat] Add option to return profile when editing profile

### LINKED ISSUE #129: Add return to profile button for edit profile forms
- add return to profile button for edit profile forms
- make sure you warn user to save their changes before leaving

BenJeau:
I know there's a lot duplicated code, but I'm not sure how to not duplicate it without redux since the parent (aka `EditProfileLayoutView.jsx`) has no information on the forms data.

I also removed the `<PageHeader />` from `EditProfileLayoutView.jsx` to put it in every child component to put the return to profile icon.

Let me know of any changes :)

closes #129


dynamic11:
We will hold this for after UAT release

PR #214, opened 2020-05-07
[Fix] Frontend linting 

mo-5:
Fixed the frontend linting errors. There is still some more errors that require more time to solve.

### DIFF
--- a/services/frontend-v3/src/routes/Secured.js
+++ b/services/frontend-v3/src/routes/Secured.js
@@ -25,7 +25,7 @@ function Secured(props) {
   const [keycloak, setKeycloak] = useState(null);
   const [redirect, setRedirect] = useState(null);
 
-  const changeLanguage = props.changeLanguage;
+  const { changeLanguage } = props;
 
   useEffect(() => {
     // Check if profile exist for the logged in user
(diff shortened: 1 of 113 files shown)


services/frontend-v3/src/routes/Secured.js
```diff
@@ -193,23 +192,22 @@ function Secured(props) {
...
+            {/** **** Logout authorized user ***** */}
             <Route
               exact
               path="/secured/logout"
               render={(routeProps) => (
                 <Logout keycloak={keycloak} {...routeProps} />
               )}
             />
-            {/****** 404 Page ******/}
+            {/** **** 404 Page ***** */}
```
dynamic11, line 203: this looks weird...no?

mo-5: Its using the multi-line comment syntax. Since its following the airbnb style standard.
/**
comment 
*/

I am not sure why we have so many "**". We could have used {/* comment */}

dynamic11: yea i agree with  {/* comment */} or just make it multi-line

PR #245, opened 2020-05-14
[Fix] White screen in edit profile

### LINKED ISSUE #216: Getting white screen of death when I click on "Talent" in side nav in Edit Profile
**Describe the bug**
Consistently getting white screen when clicking on "Talent" in side nav

**To Reproduce**
1. Go to 'Edit Profile'
2. Click on 'Talent' on side nav
3. Getting white screen of death

**Expected behavior**

Clicking on Talent should bring me to the talent section of the form

**Screenshots**
image

**Additional context**

BenJeau:
It seems that we need to have a default value for the state and the correct value (eg. not null/undefined for an array, it should just be an empty array [])

related to #216 (it may close it, but need to test it on remote dev first)

image


KateAThornley (approved):

dynamic11:
Tested and seems to have solved #216

PR #250, opened 2020-05-14
[Feat] Added persistance to search fields

### LINKED ISSUE #249: Have search fields persisted after searching
After searching for a profile, the fields are blank

I'd like to have the information that I entered still present on the page

Maybe using redux, but it's probably overkill for this. We should probably just parse the url and set that as the initial data

BenJeau:
Used queryString to get the values from the URL and formated the object according to ant design form values format

closes #249

### DIFF
--- a/services/frontend-v3/src/components/searchFilter/SearchFilter.jsx
+++ b/services/frontend-v3/src/components/searchFilter/SearchFilter.jsx
@@ -15,13 +15,17 @@ const SearchFilter = ({ history, changeLanguage }) => {
   const [branchOptions, setBranchOptions] = useState([]);
   const [locationOptions, setLocationOptions] = useState([]);
   const [classOptions, setClassOptions] = useState([]);
-  const [previousValues, setPreviousValues] = useState(null);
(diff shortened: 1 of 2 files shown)


services/frontend-v3/src/components/searchFilter/SearchFilter.jsx
```diff
@@ -14,11 +15,41 @@ const SearchFilter = ({ history, changeLanguage }) => {
   const [branchOptions, setBranchOptions] = useState([]);
   const [locationOptions, setLocationOptions] = useState([]);
   const [classOptions, setClassOptions] = useState([]);
+  const [previousValues, setPreviousValues] = useState(null);
 
   const toggle = () => {
     setExpand(!expand);
   };
 
+  const getPreviousValues = useCallback(() => {
```
dynamic11: I would rename "getPreviousValues" to something like "getSearchFeildValues" to be more descriptive and maybe add a small comment about what the function does

dynamic11 (approved):

PR #258, opened 2020-05-19
[Update] Refactor backend

mo-5:
- Changed files and separated the API folder into routes and core.
- Moved to yarn from npm to stay consistent with front-end 
- Fixed some eslint errors
- Updated routes so that they stay consistent and in one location this will make scalability much easier 
- Updated Node packages to latest 
- Removed unused Node packages


TrevorBivi:
I'm getting an error on /admin/dashboard on the frontend because growthRateByMonth.current_month_additions has been changed to growthRateByMonth.currentMonthAdditions in the backend's response for api/admin/dashboard/. (This should probably be fixed on the frontend side)

mo-5:
> I'm getting an error on /admin/dashboard on the frontend because growthRateByMonth.current_month_additions has been changed to growthRateByMonth.currentMonthAdditions in the backend's response for api/admin/dashboard/. (This should probably be fixed on the frontend side)

Good catch. I have fixed it in the front-end. Please let me know if you noticed anything else. Thanks.

TrevorBivi:
I've also noticed that the /dashboard/competencies causes the backend to get an error
TypeError: Cannot read property 'findAll' of undefined
upskill-backend |     at getOption (/upskill-backend/src/core/admin/admin.js:26:28)

This doesn't happen on skills. Looking into it a bit more

TrevorBivi:
The issue seems to be that there are a few areas in /core/admin/admin.js where getModel is being called with "competency" instead of "skill" as the parameter

mo-5:
I have made some changes to the admin file. Please let me know if you are still getting the same errors. Thank you.

TrevorBivi (approved):

PR #259, opened 2020-05-19
Fix admin dashboard infinite calls

BenJeau:
Added empty dependency array, to only run when the component mounts


KateAThornley (approved):
Phenomenal programming. Truly excellent Ben.

(thank you for fixing this lol)

PR #260, opened 2020-05-19
[Feat] Added redux for change language

### LINKED ISSUE #265: Cleanup code by using a state management tool
The code can get overwhelming when passing props multiple levels of components and the manipulation of local storage may not be ideal

A React way to handle local storage and simplify passing props, maybe something like `mobx` or `redux`

**Additional context**
Just a part of cleaning up the code

BenJeau:
This is just an initial wave of redux changes to get the feedback of devs on the team.

This PR shows that by using redux, we don't need to pass `changeLanguage` to pretty much every component and that we would not need to play with `localStorage` to get the value of the locale (`redux-persists` takes charge of persisting the state).

### Added modules
- `redux-persist` (https://github.com/rt2zz/redux-persist) - to persist the redux state in localStorage
- `@reduxjs/toolkit` (https://redux-toolkit.js.org/) - to make the use of redux way easier with slices and ability to mutate state
- `react-redux` (https://github.com/reduxjs/react-redux) - to use redux in a react project
- `redux` (https://github.com/reduxjs/redux) - to use redux

Feel free to comment or question or hate on this idea haha, let me know what you guys think

closes #265

### DIFF
--- a/services/frontend-v3/package.json
+++ b/services/frontend-v3/package.json
@@ -5,6 +5,7 @@
   "dependencies": {
     "@ant-design/compatible": "^1.0.2",
     "@ant-design/icons": "^4.0.5",
+    "@reduxjs/toolkit": "^1.3.6",
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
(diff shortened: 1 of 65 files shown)


BenJeau:
I just saw that in the code we use a lot of `intl.formatMessage({ id: "language.code" })` to get the locale, we should instead get it from redux, I'll do another commit. It will clean up the use of `injectIntl`

BenJeau:
Note: debugging redux applications may be a pain, instead of console logging everything I recommend using [Redux DevTools](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd)

TrevorBivi (approved):

PR #273, opened 2020-05-19
[Feat] Show user that edit profile forms have unsaved info

### LINKED ISSUE #128: See if profile forms have unsaved changes
- add a small status for the profile forms to see if there are unsaved changes

BenJeau:
### Notes
- `getInitialValues()` had to be moved above in the component, since it is used in `checkIfFormValuesChanged()`, and that is used in `onSave()`
- `lodash` is used to easily compare if the form values have changed and that module was already imported in the project, `QualificationsFormView.jsx` has a modified `checkIfFormValuesChanged()` since I think there's a bug with the forms in ant design (read the comment for that method to know more)

### Known bugs
- the form will say it is "unsaved" after the user clicks on the save button, I will open a separate issue to fix this. It involves changes in how form values are handled and redux will be involved.

closes #128

### DIFF
--- a/services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
+++ b/services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
@@ -20,6 +20,7 @@ import { RightOutlined, CheckOutlined } from "@ant-design/icons";
 import { FormattedMessage, injectIntl } from "react-intl";
 import axios from "axios";
 import moment from "moment";
+import _ from "lodash";
 import {
   KeyTitleOptionsPropType,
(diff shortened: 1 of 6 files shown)


dynamic11 (changes requested):
Consider making the "unsaved" text slightly smaller. maybe italic? maybe use <SaveOutlined />?
image

BenJeau:
A lighter shade and italic? (I know I don't have the right font and I'm not sure why, but my local dev has always been like that)
image

dynamic11:
> A lighter shade and italic? (I know I don't have the right font and I'm not sure why, but my local dev has always been like that)
> image

looks good will approve!

dynamic11 (approved):

PR #281, opened 2020-05-20
[Fix] Mobile navbar

### LINKED ISSUE #125: Fix UI squeeze when nav is expanded
- in mobile the sidebar auto collapses
- when it is expanded it squeezes the content of the page and it can make things look broken
- we need to find a way for it to either push the content off the page or just hid content while sidebar is expanded

**_NOTE:_** As we don't need mobile support this is low priority

BenJeau:
When the user is not logged in
image

When the user is logged in
image
image

closes #125


BenJeau:
@dynamic11  I'm not too sure about putting the Home in the hamburger menu, I thought of putting the changeLanguage in there, but it would require more work and wanted your opinion on it before I would do it.

dynamic11:
> @dynamic11 I'm not too sure about putting the Home in the hamburger menu, I thought of putting the changeLanguage in there, but it would require more work and wanted your opinion on it before I would do it.

I think what you have is good. This won't be used on mobile anyways so im not too worried

dynamic11 (approved):

PR #311, opened 2020-05-28
[Fix] unsaved indicator bug

### LINKED ISSUE #283: Edit profile showing unsaved indicator after saving form
**Describe the bug**
Edit profile showing unsaved indicator after saving form and the clear button does not work as intended after saving the form

**To Reproduce**
1. Go to 'Edit profile'
2. Enter some new information in the forms
3. Click on the 'Save' button
4. The unsaved indicator will still be there and the clear button will reset the value to the initial values

**Expected behavior**
The unsaved indicator should disappear after saving the form (clicking the save button)

**Additional context**
#273

BenJeau:
- Saves the updated values to the state and compares the form values to those instead of the initial values passed from the props
- Updates the initial values for the form to the saved values for it to reset to the saved values and not the initial values form the props (if the form has been saved) 

closes #283


BenJeau:
Just found the Qualifications forms has still the bug

dynamic11 (approved):
Looks Good!

PR #313, opened 2020-05-28
Add prettierrc file to frontend

mo-5:
I added the prettier config file to the frontend so that everyone maintains the same rules for formating


dynamic11:
can you add a reason why you are implementing this in the PR?

mo-5:
> can you add a reason why you are implementing this in the PR?

I added the prettier config file to the frontend so that everyone maintains the same rules for formating.

BenJeau (approved):

PR #340, opened 2020-06-04
 Added end points to test api

mo-5:
- api/test/  to test if the API is running 
- api/test/db to test the connection with the database
- api/test/db/users to test the user table


BenJeau (approved):
It works!

dynamic11 (approved):

PR #363, opened 2020-06-08
[New] Toggle ex feeder

### LINKED ISSUE #239: Separate ExFeeder from Talent Management Results Card
Some users may want to hide the Recent Talent Management Results card on their profile however they will most likely want to display if they're nominated in the ex-feeder program.

We need to separate the ex-feeder display from the talent management card on user profiles, or allow for the user to hide the talenet managemenr results but continue showing the ex-feeder portion of the card (i.e adding another visibility toggle perhaps)

**Additional context**
Screen Shot 2020-05-12 at 6 24 14 PM

TrevorBivi:
Adds ability to toggle ex feeder seperately from talent management

exFeeder toggle card

notExFeederCard


services/frontend-v3/src/components/layouts/profileLayout/ProfileLayoutView.jsx
```diff
@@ -330,19 +341,37 @@ const ProfileLayoutView = ({ data }) => {
           </Row>
         )}
 
-        {visibleCards.talentManagement && visibleCards.careerInterests ? (
+        {(visibleCards.talentManagement ||
+          (visibleCards.exFeeder && data.exFeeder)) &&
```
dynamic11, line 349: why are we checking both visibleCards.exFeeder && data.exFeeder

services/frontend-v3/src/components/profileCards/ProfileCardsView.jsx
```diff
@@ -92,12 +93,14 @@ function ProfileCardsView({
                 <Switch
                   checkedChildren={<EyeOutlined />}
                   unCheckedChildren={<EyeInvisibleOutlined />}
-                  checked={disabled}
+                  checked={disabled && !forceDisabled}
```
dynamic11, line 99: why do we need this?

services/backend/src/database/migrations/20191031145350-create-user-and-profile.js
```diff
@@ -152,7 +153,7 @@ module.exports = {
         });
       });
   },
-  down: (queryInterface) => {
```
dynamic11: seems to be cause by eslint. please just confirm if this styling chnage is correct (im not sure)

dynamic11 (changes requested):

services/frontend-v3/src/components/layouts/profileLayout/ProfileLayoutView.jsx
TrevorBivi: Currently on development it is not possible to show you are not an ex-feeder. If you don't check the box no message will show on your profile. I assumed this we would want to keep this because I don't think someone would ever want to publicly advertise they aren't an ex-feeder

services/frontend-v3/src/components/profileCards/ProfileCardsView.jsx
TrevorBivi: I added a message for the exfeeder card to state that the user is not an ex-feeder on the private profile so that they can be able to edit this value. I assume that we would still want to never show this information publicly so I made a way to force a card to stay hidden and I do this is they have ex-feeder unchecked

dynamic11 (approved):

PR #389, opened 2020-06-12
[Feat] Rework statistics endpoint

BenJeau:
Separated the stats endpoint from the admin endpoint
Separated the single stats endpoint to 10 different ones
Reworked some logic in those endpoints
Removed inline SQL


mo-5 (approved):
Great job on this. Really good work !

PR #398, opened 2020-06-16
[Fix] Backend core profile fix

mo-5:
- Updated how private and public profiles are handled. 
- Fixed the update method for profile.

### DIFF
--- a/services/backend/.eslintrc.json
+++ b/services/backend/.eslintrc.json
@@ -1,18 +1,19 @@
 {
-	"extends": ["airbnb-base", "prettier"],
-	"plugins": ["prettier"],
-	"rules": {
-		"prettier/prettier": [
-			"error",
-			{
-				"endOfLine": "auto"
-			}
-		],
-		"no-unused-vars": "warn",
-		"no-console": "off",
-		"func-names": "off",
-		"no-process-exit": "off",
-		"object-shorthand": "off",
-		"class-methods-use-this": "off"
-	}
+  "extends": ["airbnb-base", "prettier"],
+  "plugins": ["prettier"],
+  "rules": {
+    "prettier/prettier": [
+      "error",
+      {
+        "endOfLine": "auto"
+      }
+    ],
+    "no-unused-vars": "warn",
+    "no-console": "off",
+    "func-names": "off",
+    "no-process-exit": "off",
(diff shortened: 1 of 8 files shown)


BenJeau (commented):
Seems to work, but I think the information is returning is not the correct one

ex: skills
image
should probably only be containing 
```json
{
  "id": "",
  "name": ""
}
```
and same thing with competencies, developmentalGoals and for many of the different fields.

I think we should not send the updatedAt/createdAt to the client and also if the id is not relevant, like the one in visibleCards (if we want to update that info, it should just fetch the one of the user and update that visibleCard)

BenJeau:
And we should only send back the info in one of the language 

ex:
image

to become?
```json
{
  "tenure": "Term",
  "careerMobility": "Develop in role",
  "jobTitle": "Manager",
  "branch": "Human Resource Branch",
}
```

BenJeau:
Looks way better than before, but just one comment, I made a query to `/api/profile/:id`
image
I think for these values, since its not a list, I would just return something like 
```json
{
    "groupLevel": "CS 05",
    "actingLevel": null,
    "securityClearance": "Top Secret",
    "lookingJob": "Not looking",
    "tenure": "Term",
    "careerMobility":  "Develop in role",
    "jobTitle": "Manager",
    "branch": "Human Resources Branch",
    "talentMatrixResult":  "Early promise"
}
```
what do you think?

BenJeau:
Also I don't see you returning info about the second language proficiencies, in the past we were returning the info like this 
image
but I think it would either just be this
```js
secondLanguageProficiencies: {
  oral: {
     level: 'A',
     date: '',
  },
  writing: {
     level: 'A',
     date: '',
  },
  reading: {
     level: 'A',
     date: '',
  },
}
```
or like this since its stored as an array in the db

```js
secondLanguageProficiencies: [
  {
     proficiency: "oral",
     level: 'A',
     date: '',
  },
  {
     proficiency: "writing",
     level: 'A',
     date: '',
  },
  {
     proficiency: "reading",
     level: 'A',
     date: '',
  },
]
```

mo-5:
Fixed in the new commit. I also added ID check to make sure the user matches for the private profile. I also fixed the frontend for creating user in the database.

BenJeau (approved):

PR #405, opened 2020-06-17
[Fix] Make admin dashboard work with new backend api

### LINKED ISSUE #238: Admin Dashboard "Popular Skills" graph displays as empty
**Describe the bug**
On the admin dashboard the Popular Skills graph is empty even when users have skills on their profiles. There should be a bargraph display of skills most frequently selected for users' profiles.

**To Reproduce**
1. Go to Admin Dashboard dev or uat link of I-Talent
2. Should see an empty graph under "Popular Skills" header, while other graphs display information properly.

**Screenshots**
Screen Shot 2020-05-12 at 2 47 49 PM

BenJeau:
There's a lot of changes in this PR, I would recommend you to either look by commit or just test it.

### Overall goal
Make the admin part of the UI work with the new backend (changed every API call to the backend and the way that the data is formatted)

### UI Changes
- Changed how data is stored after we make a call (it now stores it in redux, and uses it as a cache)
- Now the tables are showing the loading indicator instead of blanking out the page
- Since the cards in the dashboard now have separate endpoints, they can now load on their own and I just added the loading prop to them
- Cleaned up the components (removed useless functions, try to fix eslint warnings, used `<FormattedMessage />` instead of `intl.formatMessage`, cleaned up props and PropTypes)
- Added more descriptive message when deleting entries (since they now also delete any relation in the db - ex. deleting a skill will consequently delete the skills and developmentalGoals associated to profiles)
- Set the document title only once, instead of on every render
- When cancelling a modal, it now shows an info prompt instead of an error one

### API changes
- Added the createdAt field to return when querying `/api/admin/users` (the table displays when the user was registered/created)
- Changed how schools are updated (the translations were not saving since if there was only one translation, like a French one, and you'd want to add an English one, you would need to create a new translation and not update one)
- The `/api/option/skillsAllLang` also returns the `categoryId` to match with the category information that is fetched
- Removed `/api/option/categoriesSkills` since you can just query the categories and skills independently (which may be better here since we may already have the info requested when we looked at the categories page, and it was maybe cached in redux)
- Fixed the stats endpoint (added / for the count endpoint in stats)

### Fix
This also fixes #238


mo-5 (approved):
I have tested it locally and it works

PR #413, opened 2020-06-19
[Fix] Fixed forms backend

BenJeau:
### Changes
- Fixed the update profile endpoint (for it to create, delete and update row in the tables respectfully)
- Added signup step in the db (this is now used to keep track users that have completed the signup flow)
- Made some columns optional and changed how the unique property worked for developmental goals (these changes were made because of the update of the profile endpoint)
- Fixed the profile forms
- Added validation to the update profile endpoint


BenJeau:
@MohamedRadwan  I'm just having merge conflicts because of this, just curious, why did you add all of the new fields? https://github.com/CDH-Studio/UpSkill/pull/412/files#diff-c75e7985a683c76ac1a873f2289a2555 if it was intentional, you need to create a migration, or update the existing one

mo-5:
> 
> 
> @MohamedRadwan I'm just having merge conflicts because of this, just curious, why did you add all of the new fields? https://github.com/CDH-Studio/UpSkill/pull/412/files#diff-c75e7985a683c76ac1a873f2289a2555 if it was intentional, you need to create a migration, or update the existing one

I didn't actually add anything to the database I think prettier automatically formated the file. You can just override the changes in schema.prisma since I didn't modify any of the core content/logic.

BenJeau:
Oh weird, your pretty actually formats the prisma file? Mine doesn't, so maybe the changes are good. How did you make it work for the prisma file?

BenJeau:
PR should be good to be checked, please try to break it or find some edge cases which I may have not caught

services/backend/src/core/user/user.js
```diff
@@ -54,55 +58,82 @@ async function getUserById(request, response) {
...
     console.log(error);
+    if (error.errors) {
+      response.status(422).json(error.errors);
+      return;
+    }
     response.status(500).json("Unable to create profiles");
   }
 }
 
 async function checkExistence(request, response) {
```
mo-5: I think this method will become useless since we have a sign up step created as soon as the user signs up. What do you think?

services/backend/src/router/profile/profile.js
```diff
@@ -13,11 +17,7 @@ profileRouter
     [UUIDValidator, langValidator],
     profile.getPublicProfileById
   )
-  .put(keycloak.protect(), profile.updateProfile);
-
-profileRouter
-  .route("/:id/status")
-  .get(keycloak.protect(), UUIDValidator, profile.getProfileStatusById);
+  .put(keycloak.protect(), updateProfileValidator, profile.updateProfile);
```
mo-5, line 20: we should also add UUID validation for the put request.

services/backend/src/router/user/user.js
```diff
@@ -1,16 +1,22 @@
...
 const userRouter = Router();
 
 userRouter.get("/", keycloak.protect(), user.getUsers);
 
 userRouter
   .route("/:id")
-  .get(keycloak.protect(), user.getUserById)
-  .post(keycloak.protect(), user.createUser);
+  .get(keycloak.protect(), [UUIDValidator], user.getUserById)
+  .post(keycloak.protect(), createUserValidator, user.createUser);
```
mo-5, line 13: Should we check for UUID in the post ?

services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataForm.jsx
```diff
@@ -29,64 +29,47 @@ const EmploymentDataForm = ({ formType }) => {
...
-        title: result.data[i].description[locale],
-        key: result.data[i].id,
-      };
-      options.push(option);
-    }
-    setSubstantiveOptions(options);
-    return 1;
+    const result = await axios.get(
+      `${backendAddress}api/option/tenures?language=${
+        locale === "en" ? "ENGLISH" : "FRENCH"
```
mo-5, line 34: I noticed that we setting the locale to "ENGLISH" and "FRENCH" or are we still using en and fr ?

mo-5: So I double checked and the language is locale seems to be stored as en. I wonder if its better to change either the frontend or the backend so we stay consistant

mo-5:
I think we should add validation for the phone numbers in the update profile method ( that they are in the correct format). As a backup to the frontend validation.

services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataForm.jsx
BenJeau: I think it would be easier to change it in the frontend since we are just setting it in one place, and if we were to change the backend to accept fr or en, we would need to change quite a few things (I think) should the change be in this pr?

services/backend/src/router/user/user.js
BenJeau: the createUserValidator is an array including the UUIDValidator, should I leave it out of createUserValidator?

services/backend/src/router/profile/profile.js
BenJeau: same as below

services/backend/src/core/user/user.js
BenJeau: yeah, I think so too, I'm not querying that endpoint while signing up, I just do a get, and catch it if it throws a 500, the do a post

mo-5:
After I create my profile and click view profile I get this error 
image

BenJeau:
I didn't fix the profile view yet, I'm doing that right now

BenJeau:
But edit profile should work

services/backend/src/router/user/user.js
mo-5: I feel like we should separate them since UUID validation is done almost on all profiles and user related requests so we can write one method and share it for all of them. That way it would be easier to maintain.

services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataForm.jsx
mo-5: I think a frontend change would be better since most of the backend is built to use the full word. If the frontend change is minor I think we could do it in this pull request so its all done at once.

mo-5:
> 
> 
> But edit profile should work

I have tested the edit and create new profile and it works

services/backend/src/router/user/user.js
BenJeau: but createUserValidator is just an array, containing the UUIDValidator variable, not redeclaring the UUID body check

mo-5 (approved):

PR #414, opened 2020-06-22
[Feat] Added basic documentation with swagger

BenJeau:
Added docs to four routes with swagger, just to test it out and set it up

Introduces three new env variables for the backend for swagger to query the API
- `DOCS_KEYCLOAK_TOKEN_URL`
- `DOCS_KEYCLOAK_AUTH_URL`
- `DOCS_KEYCLOAK_TOKEN_URL`

**Note:** This may start to bloat the routes now that we'll add jsdoc comments for every endpoint, maybe we should separate the routes? not sure and for sure up for discussion

Runs locally on http://localhost:8080/api-docs and looks something like this
image


mo-5 (approved):

mo-5:
I do agree about the router folder getting messy speical in something like options. I am not sure if there is a better option to go with.

BenJeau:
Another way to document with swagger is to just have separate .yaml or .json files instead of having those comments. I initially thought comments would be better since we would be more prone to update those when you update the code, but now I'm not sure. Either way, the comments need to be formatted like a .yaml file (indentation wise)

mo-5:
Interesting. I see your point about it not getting updated as code will get updated but I feel like the more code we write the harder it will be to read and modify the code. We can keep it in a sperate file and ask everyone to modify it and make sure it looked at in the code reviews.

PR #423, opened 2020-06-23
[Fix] Fixed profile view

BenJeau:
Reduced the number of calls the profile view makes to the backend
Fixed the API calls made to the backend (including how the data is formatted)
Make skills and mentorshipSkills return the categoryId and category name in the profile response

### DIFF
--- a/services/backend/src/core/admin/util/users.js
+++ b/services/backend/src/core/admin/util/users.js
@@ -47,8 +47,8 @@ const getUsers = async (request, response) => {
     });
 
     const users = usersQuery.map((i) => {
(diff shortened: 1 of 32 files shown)


mo-5:
I am getting this error on admin users page. I think this is caused becuase we removed the get all users method. 

image

mo-5:
I get this error when I try to visit mary doe or john does profile 
image

Looks like the issue is the api call is not made with the language
image

BenJeau:
The bugs you encountered should be fixed now

mo-5 (approved):
Looks good !

PR #424, opened 2020-06-23
[Fix] Renamed locale frontend to ENGLISH and FRENCH

BenJeau:
This will simplify calls to the backend

This PR also contains other fixes


mo-5 (approved):
Tested it locally and it works

PR #431, opened 2020-06-25
[Fix] Catch empty search and do not allow user to search

### LINKED ISSUE #421: Search should not allow user to search with all fields empty
**Describe the bug**
When clicking "search" without entereing information into any of the fields, there is no validation to tell the user to search for something, instead the user is just lead to the results page with a text showing a query error.

**To Reproduce**
1. Go to home page of I-Talent
2. Click on "Search" button
3. See text displayed "An error was encountered! Please try again. Error: invalid query"

**Expected behavior**
There should be validation on the home screen telling the user to either input terms into the basic search or the advanced search (depending on the search form the user is using) when the user tries to search with no search criteria entered.

**Screenshots**
Screen Shot 2020-06-23 at 11 05 49 AM

KateAThornley:
- added ternary operator to catch empty query (no search values entered) and dont push to results page until values have been entered

### DIFF
--- a/services/frontend-v3/src/components/searchBar/SearchBar.jsx
+++ b/services/frontend-v3/src/components/searchBar/SearchBar.jsx
@@ -55,10 +55,10 @@ const SearchBar = () => {
     const query = queryString.stringify(values, { arrayFormat: "bracket" });
     const url = `/secured/results?${encodeURI(query)}`;
     {
-      query !== ""
-        ? history.push(url)
-        : // eslint-disable-next-line no-console
-          console.log("");
+      if (query !== "") {
+        history.push(url);
+        // eslint-disable-next-line no-console
+      } else console.log("");
     }
   };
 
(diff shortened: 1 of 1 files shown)


services/frontend-v3/src/components/searchBar/SearchBar.jsx
```diff
@@ -52,7 +54,12 @@ const SearchBar = () => {
   const handleSearch = (values) => {
     const query = queryString.stringify(values, { arrayFormat: "bracket" });
     const url = `/secured/results?${encodeURI(query)}`;
-    history.push(url);
+    {
+      query !== ""
+        ? history.push(url)
+        : // eslint-disable-next-line no-console
+          console.log("");
+    }
```
BenJeau: I would change the ternary operator for a plain old if statement here, making it more readable (and maybe you won't need the eslint disable at the top of the file?)
```suggestion
    if (query !== "") {
      history.push(url);
    }
```
Appart from that, looks good

BenJeau (approved):

BenJeau:
Just curious, why do you have a console.log in the else?

BenJeau:
Another small comment, is the {} surrounding the if statement necessary?

PR #450, opened 2020-07-03
[Fix] openshift script and added new example envs

BenJeau:
Openshift didn't like not having the database client


mo-5 (approved):

BenJeau:
I somehow messed up this branch, I'll just push it directly to `backend-change`

* the last commit

PR #457, opened 2020-07-03
[New] Update user profile when the lanauge is changed 

### LINKED ISSUE #157: Add preferred language in settings for user profile
**Is your feature request related to a problem?
The user has to change the language every time the session ends. In the long term this will become frustrating for users. 

As the user to input preferred language when they create their profile once the profile is created the language is stored in the db. The user will then have the ability to change languages from settings.

Storing the language in local storage would be the only useful alternative but it could cause issues to users.

mo-5:
- Updated user profile once the language is changed so that the user doesn't have to change their language every time they log in. 

closes #157

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+REDIS_PASSWORD=[can be anything you want]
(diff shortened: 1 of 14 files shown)


BenJeau:
Looks good, can you also make it so that when we get the user in Secured.tsx to get that info from the API (if it's not already done) and change the locale in redux based on that?

mo-5:
Should be good now

BenJeau (approved):

PR #458, opened 2020-07-03
[Feat] Added redis connect for session management

### LINKED ISSUE #447: MemoryStore in production should not be used
In the logs of openshift, this error is shown and should be fixed.

image

Per discussion with @dynamic11, Redis connect seems like a good solution https://www.npmjs.com/package/connect-redis

BenJeau:
### Changes
- Changed based docker images to all use the alpine version (which is smaller, and less vulnerable) ex: 
_before_ >
MicrosoftTeams-image (1)
_after_ >
image
- Added redis session management
- Properly using volumes and networks (instead of needlessly exposing ports)

closes #447

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,7 +25,7 @@ services:
 
   db-postgres:
     container_name: "upskill-database"
-    image: postgres
+    image: "postgres:12-alpine"
     restart: always
     environment:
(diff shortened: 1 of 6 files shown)


BenJeau:
Note: will need to update openshift with a new redis pod

mo-5 (approved):

Issue #460, opened 2020-07-03
Show code coverage in PR


BenJeau:
To have good code, we need a good code coverage. Making devs realize the code coverage in PRs is a great way to have good coverage and having tests for the code.

- Add a github bot for test coverage or something similar

mo-5:
This looks like a good option 
https://github.com/apps/codecov

BenJeau:
Wonder what ISED is using for that, seems like they are only free if we are opensource, which we are for now, which is good, but I think they want to make it private in the future. Meanwhile, I'll just add one of those tools. @MenaMachado is there a code coverage tool that ISED uses?

BenJeau:
Will probably be done by the CI/CD team

PR #465, opened 2020-07-06
[Feat] Added github Action backend test

### LINKED ISSUE #459: Backend tests should be run on every PR
Since we have backend tests, they should be used to check if PRs have broken the backend or not.

We should add a new GitHub action workflow with that job.

BenJeau:
- Added github action test
- Added new backend environment variable `REDIS_HOST` 

closes #459

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,6 +2,9 @@ name: Backend tests
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


BenJeau:
Once this PR is merged, I will make the backend test as Required and implement this #460

mo-5 (approved):

PR #472, opened 2020-07-07
[Feat] Added sonarqube github action

BenJeau:


mo-5:
should we separate sonarqube scan into another action file? Oh didn't notice this was a draft

BenJeau:
not exactly sure, since I think it hooks into the coverage, so we wouldn't need codecov

BenJeau:
will close since its not a priority and seems to be something the CI/CD team has in mind

PR #507, opened 2020-07-09
[Fix] Admin router

### LINKED ISSUE #464: If user was in the admin section, the user interface will not say that the user is logged in
**Describe the bug**
If user was in the admin section **and they need to log back in**, the user interface will not say that the user is logged in

**To Reproduce**
1. Go to any admin page
2. **Be timed out**
3. Log back in

**Expected behavior**
Admin routes should make the same requests as the secured routes. The content in local storage should be updated and it should make some requests to the backend to see if the user exists.

**Screenshots**
The sign in button should not be shown
image

BenJeau:
- Removed admin check route
- Properly check if the user has permission for admin route and render Forbidden component if they don't
- Clean up the routes files
- Centered the content in the error pages
- Removed sentry in the frontend
- Added icons tot he buttons in error screens (just saw ant-design was doing that and makes it less intimidating, at least for the error page)
image
image
image

fixes #464


mo-5:
I like the changes but I am not sure about the wording of the 404 and 403. I feel like for a government website we should just keep it simple: "Access denied" etc.

mo-5 (approved):

BenJeau:
Yeah, it was already like that in the code, I admit its kinda weird haha

PR #517, opened 2020-07-10
[Fix] Admin check

BenJeau:
Should fix the admin check in the frontend. Cannot test right now as I don't have admin privileges for the application.


mo-5 (approved):
tested locally and works

PR #534, opened 2020-07-14
[Feat] Add i18n checks and add pre-commit hooks

### LINKED ISSUE #466: Check i18n files for missing and unused keys
**Note:** This would be a nice to have to give us peace of mind knowing that the app is translated

The following could be done in a GitHub action or just in a script in package.json
- To make sure everything is in both languages
  - in the i18n JSON files, they should both have the same keys (simple javascript)
  - detect hardcoded strings in the app (could extend eslint https://www.npmjs.com/package/eslint-plugin-react-intl)
- To clean up the files
  - we could automatically remove unused keys by searching the whole frontend project to see if they are used (simple javascript)
  - we could sort the whole file (lodash/javascript)

Would prevent issues like #417 #378 #211 #151

BenJeau:
### Changes
- Cleaned up i18n files
- Added 3 pre-commit hooks
  - i18n lint
  - frontend lint
  - backend lint
- There's a downfall, you'll need to run `yarn` at the root of the project for the dependencies of pre-commit to install and set in the .git folder

### Screenshot
A failing check
image
A passing check
image
A cleanup
image

fixes #466


mo-5:
I like this but I was wondering if we should add linting to commits. If someone needs to commit their files temporarily or needs help and someone wants to access the latest of the branch they wouldn't be able to commit. Do you know what I mean?

BenJeau:
Yeah, I understand, so the thing is you can always bypass it (if you want). Example of it failing
image

The only downside is I'm not sure if it's supported by integrated VCS in IDEs, like git in vs code

mo-5 (approved):

mo-5:
Just wondering can we add the precommits to each service? like the backend and frontend? or does it have to be in the root dir ?

BenJeau:
Since pre-commit hooks are linked to the .git folder which is at the root of this project, it needs to be at a project level, if our setup would be split up in two git repos, we could.

mo-5:
Do you want to run the script through and push these changes?

mo-5:
https://github.com/magicmark/pre-commit-vscode

Could you check this out? We can run precommits straight from vscode. I am not sure if this will be better/easier to set up for developers.

BenJeau:
I think I'll just omit the pre-commit hooks for now (since they require you to have node_modules installed to push - which you should, but may not be the case if you strickly use Docker) and I'll pull the latest i18n files and run the script. I kinda forgot about this PR haha

Issue #491, opened 2020-07-08
Grey out cards when set to private group instead of displaying no infromation


mo-5:
When a user sets a card to a private group should be greyed out instead of displaying no information. 

![image](https://user-images.githubusercontent.com/21133129/86977164-02fb8480-c14a-11ea-92b7-4a17d561fe04.png)
![image](https://user-images.githubusercontent.com/21133129/86977203-1e668f80-c14a-11ea-9772-17ad170f424e.png)

Issue #490, opened 2020-07-08
Some cards are visible when set to private group 


mo-5:
The following cards are visible when set to the private group:
- Recent talent management results
- Job mobility
- Mentorship skills
- Employee summary

The rest of the cards are working properly.

![image](https://user-images.githubusercontent.com/21133129/86977255-3b9b5e00-c14a-11ea-8047-bc4aae6a09fb.png)
![image](https://user-images.githubusercontent.com/21133129/86977269-40f8a880-c14a-11ea-9f04-5243412c4342.png)
![image](https://user-images.githubusercontent.com/21133129/86977281-4655f300-c14a-11ea-90c8-cb9933a6eb18.png)

Issue #484, opened 2020-07-08
Friends should be called "Private group"


BenJeau:
**Is your feature request related to a problem? Please describe.**
It just makes more sense to have it called "Private group", since friends implies a two-way relationship

**Describe the solution you'd like**
Rename where we use friends to use "Private group"

**Additional context**
This is from a conversation with Mena

MenaMachado:
Hi guys, I was speaking with Ali, and he suggested calling them "connections", which I like. So basically we are adding connections to our private group.

MenaMachado:
Perhaps, the Heading is called Private Group but the subheading "Connections"

![image](https://user-images.githubusercontent.com/49656364/86970847-1bfe3880-c13e-11ea-99b8-8bcd1683e58e.png)

MenaMachado:
For the privacy button, instead of "friends only", can you change it to "private group only". It implies the entire group of connections can see the info.  You may have already thought of this, we should change "Add Friend" to "Add connection". Connection needs to be lower case. 
![image](https://user-images.githubusercontent.com/49656364/86974884-16582100-c145-11ea-9fef-9dd667e4d5c0.png)

Issue #483, opened 2020-07-08
Cards should not be visible to the public by default


BenJeau:
**Is your feature request related to a problem? Please describe.**
The cards in the profiles should be seen by everyone by default when creating a new profile

**Describe the solution you'd like**
Set the cards to private by default

**Additional context**
This is from a conversation with Mena

PR #538, opened 2020-07-15
[Fix] visible cards

mo-5:
- Profile view and controller
 - Change from friends to connections
 - Cards private by default

Potentially fixed all the issues with the connection group. Needs to be tested once deployed to development.

issues #490, #491, #484, #483


services/backend/src/core/profile/profile.js
```diff
@@ -1044,96 +1044,103 @@ async function getPublicProfileById(request, response) {
...
-      }
+      const isConnection = result.connections.some(
+        (item) => item.id === userId
+      );
 
       if (
         result.visibleCards.info === "PRIVATE" ||
-        (result.visibleCards.info === "FRIENDS" && !isFriends)
+        (result.visibleCards.info === "CONNECTIONS" && !isConnection)
       ) {
```
BenJeau: For all of these big if statements, since they all have the "same" logic, we could create a function that just receives the visibleCards key

something like this just above the first if statement?
```js
const hideCard = (key) => result.visibleCards[key] === "PRIVATE" || (result.visibleCards[key] === "CONNECTIONS" && !isConnection);
```
Then the if statements could look a bit cleaner
```js
if (hideCard("info")) { .... }
```

services/backend/src/core/search/util/getAllProfiles.js
```diff
@@ -33,44 +33,49 @@ async function getAllUsers(searchValue, language, userId) {
...
+        );
 
         return {
           id,
           visibleCards: {
-            info: !(info === "PRIVATE" || (info === "FRIENDS" && !isFriends)),
+            info: !(
+              info === "PRIVATE" ||
+              (info === "CONNECTIONS" && !isConnection)
+            ),
```
BenJeau: Something similar to the other comment, maybe export that logic into another function

```js
const visibleCardBool = (value) => !(value === "PRIVATE" || (value === "CONNECTIONS" && !isConnection));
```

Then the return would just be
```js
info: visibleCardBool(info),
```

If you want, you could approach it the same way as the other comment, with keys, you would just not decompose the values of `visibleCards` in the map.

services/frontend-v3/src/components/connections/ConnectionsView.jsx
```diff
@@ -2,14 +2,14 @@ import React from "react";
 import PropTypes from "prop-types";
 import { Row, Col, List } from "antd";
 
-const FriendsView = ({ friends }) => {
+const ConnectionsView = ({ Connections }) => {
```
BenJeau: props should be camelCased
```suggestion
const ConnectionsView = ({ connections }) => {
```

services/frontend-v3/src/components/connections/ConnectionsView.jsx
```diff
@@ -2,14 +2,14 @@ import React from "react";
...
   return (
     <Row>
       <Col xs={24} lg={24}>
         <List
           size="small"
           itemLayout="horizontal"
-          dataSource={friends}
+          dataSource={Connections}
```
BenJeau: same here

services/frontend-v3/src/components/connections/ConnectionsView.jsx
```diff
@@ -23,10 +23,10 @@ const FriendsView = ({ friends }) => {
   );
 };
 
-FriendsView.propTypes = {
-  friends: PropTypes.arrayOf(
+ConnectionsView.propTypes = {
+  Connections: PropTypes.arrayOf(
```
BenJeau: same here

services/frontend-v3/src/components/connections/Connections.jsx
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import PropTypes from "prop-types";
+import ConnectionsView from "./ConnectionsView";
+import ProfileCards from "../profileCards/ProfileCards";
+
+const Connections = ({ data, title, cardName, id, type, visible, editUrl }) => {
+  return (
+    <ProfileCards
+      title={title}
+      content={<ConnectionsView Connections={data.connections} />}
```
BenJeau: props should be camelCased

services/frontend-v3/src/components/careerInterests/CareerInterestsView.jsx
```diff
@@ -11,15 +11,18 @@ const CareerInterestsView = ({ info, relocationLocationsInfo }) => {
         dataSource={dataSource}
         renderItem={(item) => (
           <List.Item>
-            <List.Item.Meta title={item.title} description={item.description} />
+            <List.Item.Meta
+              title={item.title ? item.title : null}
+              description={item.description ? item.description : null}
```
BenJeau: if the prop is not defined, we should pass undefined instead of null. So returning undefined here comes back to the same code as before. Here's one reason why https://stackoverflow.com/questions/34809178/react-js-default-prop-is-not-used-with-null-is-passed. Unless there's a reason why you are passing null here

services/frontend-v3/src/components/employeeSummary/officialLanguage/OfficialLanguage.jsx
```diff
@@ -22,43 +22,31 @@ const OfficialLanguage = ({ data }) => {
   };
 
   const getSecondLanguageGradeInfo = (dataSource) => {
-    const reading = dataSource.secondLangProfs.find(
-      (i) => i.proficiency === "READING"
-    );
+    const reading = dataSource.secondLangProfs
+      ? dataSource.secondLangProfs.find((i) => i.proficiency === "READING")
+      : null;
```
BenJeau: To keep consistency, I would return `undefined` instead of `null`, since if the `.find()` function doesn't find something that satisfies the expression inside the function, it would return `undefined`. Same thing for the other paces below
```suggestion
      : undefined;
```

services/frontend-v3/src/components/profileCards/ProfileCards.jsx
```diff
@@ -1,21 +1,66 @@
...
+    const { visibleCards } = data;
+    const modifiedCard = cardName;
+    visibleCards[modifiedCard] = value;
+    await axios
+      .put(`api/profile/${urlID}?language=${locale}`, {
+        visibleCards,
+      })
+      .catch((error) => handleError(error, "message"));
+    getCardStatus();
+  };
```
BenJeau, line 45: Love the cleanup of the view :heart:

services/frontend-v3/src/i18n/en_CA.json
```diff
@@ -274,6 +273,7 @@
   "profile.email": "Email",
   "profile.ex.feeder": "I am nominated in the EX-feeder program",
   "profile.not.ex.feeder": "I'm not nominated in the EX-feeder program.",
+  "profile.private.ex.feeder": "Ex-Feeder is set to private",
```
BenJeau: To keep consistency

```suggestion
  "profile.private.ex.feeder": "EX-feeder is set to private",
```

services/frontend-v3/src/i18n/en_CA.json
```diff
@@ -385,12 +383,14 @@
   "sign.out": "Logout",
   "home": "Home",
 
-  "profile.addFriend": "Add Friend",
-  "profile.removeFriend": "Remove Friend",
-  "profile.removeFriend.confirm": "Are you sure you want to remove this user from your friend group ?",
+  "profile.addConnection": "Add connection",
+  "profile.removeConnection": "Remove connection",
+
+  "profile.privateGroup": "Private Group",
```
BenJeau: Make it lowercase https://github.com/CDH-Studio/UpSkill/pull/267#issuecomment-634277631
```suggestion
  "profile.privateGroup": "Private group",
```

services/frontend-v3/src/i18n/fr_CA.json
```diff
@@ -276,6 +275,7 @@
   "profile.email": "Courriel",
   "profile.ex.feeder": "Je suis nommé dans le programme EX-feeder",
   "profile.not.ex.feeder": "Je ne suis pas nominé dans le programme EX-feeder",
+  "profile.private.ex.feeder": "Ex-Feeder est défini sur privé",
```
BenJeau: To keep it consistant and french translation change
```suggestion
  "profile.private.ex.feeder": "Le résultat d'EX-feeder est privé",
```

services/frontend-v3/src/pages/Profile.jsx
```diff
@@ -9,66 +9,124 @@ const Profile = ({ history, match }) => {
...
+                ),
+                experience: !(
+                  experience === "PRIVATE" ||
+                  (experience === "CONNECTIONS" && !isConnection)
+                ),
+                exFeeder: !(
+                  exFeeder === "PRIVATE" ||
+                  (exFeeder === "CONNECTIONS" && !isConnection)
+                ),
+              };
```
BenJeau: Potentially do this on the server? I saw something similar in our APIs, aren't we already doing this? If you want to keep it on the client, I would suggest the same thing as above, to extract the logic to a function

services/frontend-v3/src/i18n/fr_CA.json
```diff
@@ -388,18 +386,20 @@
...
+  "profile.removeConnection": "Supprimer la connexion",
+
+  "profile.privateGroup": "Groupe privé",
+  "profile.connections": "Connexions",
+
   "profile.visibleCards.public": "Publique",
-  "profile.visibleCards.friends": "Amis seulement",
-  "profile.visibleCards.private": "Privée",
+  "profile.visibleCards.connections": "Connexions",
+  "profile.visibleCards.private": "privée",
```
BenJeau: Not sure why this was changed to lowercase?
```suggestion
  "profile.visibleCards.private": "Privée",
```

services/frontend-v3/src/i18n/en_CA.json
```diff
@@ -385,12 +383,14 @@
   "sign.out": "Logout",
   "home": "Home",
 
-  "profile.addFriend": "Add Friend",
-  "profile.removeFriend": "Remove Friend",
-  "profile.removeFriend.confirm": "Are you sure you want to remove this user from your friend group ?",
+  "profile.addConnection": "Add connection",
+  "profile.removeConnection": "Remove connection",
```
BenJeau: This info is already defined, aka the "search.results.cards.add.connection" and "search.results.cards.remove.connection" keys, could we somehow unite those keys

services/frontend-v3/src/i18n/fr_CA.json
```diff
@@ -388,18 +386,20 @@
...
-  "profile.removeFriend": "Supprimer un ami",
-  "profile.removeFriend.confirm": "Voulez-vous vraiment supprimer cet utilisateur de votre groupe d'amis ?",
-
   "profile.is.ongoing": "Courant",
 
   "sign.out": "Déconnexion",
   "home": "Acceuil",
 
+  "profile.addConnection": "Ajouter une connexion",
+  "profile.removeConnection": "Supprimer la connexion",
```
BenJeau: same as above, and I would keep the french translations of "search.results.cards.add.connection" and "search.results.cards.remove.connection"

BenJeau (commented):
I got some weird things happening locally, even tho the fake users have all PUBLIC fields for the visible cards, I get this
image

Also, I don't think you've modified the docs for profiles, ex. it now returns connections (should update its return values (which is the `GetProfile` object in the definitions doc file).

Another overall thing with the `<ProfileCards />` instead of passing the children in the `title` prop, we should just pass it like a normal child and get the children via the children prop of ProfileCards (https://reactjs.org/docs/composition-vs-inheritance.html#containment). Ex.
```jsx
<ProfileCards
  content={<Child />}
/>
```
VS
```jsx
<ProfileCards>
  <Child />
</ProfileCards>
```
And the same would apply to the `<ProfileCardsView />`. I know this is not related to your code but I just saw it.

But good cleanup otherwise, the code looks good!

BenJeau:
Also, another small comment, the "extras" in the card does not look vertically centered
image

services/frontend-v3/src/components/careerInterests/CareerInterests.jsx
```diff
@@ -2,47 +2,62 @@ import React from "react";
...
 import CareerInterestsView from "./CareerInterestsView";
+import ProfileCards from "../profileCards/ProfileCards";
+import { ProfileInfoPropType } from "../../customPropTypes";
 
-const CareerInterests = ({ data }) => {
+const CareerInterests = ({ data, type }) => {
   const getCareerInterestsInfo = () => {
+    let description = "-";
+    if (data.interestedInRemote)
+      if (data.interestedInRemote) {
```
BenJeau: Not sure if there's supposed to be two ifs

services/frontend-v3/src/components/careerInterests/CareerInterests.jsx
```diff
@@ -2,47 +2,62 @@ import React from "react";
...
 CareerInterests.propTypes = {
-  data: PropTypes.shape({
-    interestedInRemote: PropTypes.bool,
-    lookingJob: PropTypes.shape({
-      description: PropTypes.string,
-    }),
-    relocationLocations: PropTypes.any,
-  }).isRequired,
+  data: ProfileInfoPropType,
+
```
BenJeau: remove empty line in object

services/frontend-v3/src/components/careerInterests/CareerInterestsView.jsx
```diff
@@ -11,15 +11,18 @@ const CareerInterestsView = ({ info, relocationLocationsInfo }) => {
         dataSource={dataSource}
         renderItem={(item) => (
           <List.Item>
-            <List.Item.Meta title={item.title} description={item.description} />
+            <List.Item.Meta
+              title={item.title || undefined}
+              description={item.description || undefined}
+            />
```
BenJeau: I would just revert to how it was, since if item.title is undefined, the value of title will be undefined, we don't need to specify undefined
```suggestion
            <List.Item.Meta
              title={item.title}
              description={item.description}
            />
```

BenJeau (commented):
Seems good and works well locally! Just few minor comments, then I'll approve it

BenJeau (approved):

Issue #448, opened 2020-07-03
API calls in the frontend should be standardized (using redux-thunk)


BenJeau:
### Current situation
Right now the components are making their own API calls to get data, and it may get messy passing the data to children components and making sure that the component hierarchy is structured well for passing the data down to the children. Also, we wouldn't want every child to make separate API calls, which would tax the server. In addition, this is also bloating our components with `axios` and the backend address import and taking into account the errors that the API may return scattered across the app/components.

### Admin implementation
In the new backend change, while I was adapting the admin side of the application, I was making the API calls in the controller component and saving it to redux. Children components would subscribe to redux and get the information. But to get a good flow, I was also saving in redux if the data was loading and wasn't even saving in redux if there was an error, so the children component was not aware of an error (if one would arise).

### [Redux-thunk](https://redux-toolkit.js.org/api/createAsyncThunk) (the ultimate solution, in my opinion)
We can make asynchronous calls in redux with `redux-thunk` and since we are using `redux-toolkit` it is already built-in and way easier than using `redux-thunk` barebones. With `redux-thunk` the API calls would be made in redux, making the API calls central. It also automatically creates three action types/things the components can subscribe to, either `pending`, `fulfilled`, and `rejected`. To make the call, it would be similar to the custom admin implementation, where one component calls the API, but would instead dispatch an action to redux, which would in return call the API and the components wanting that information would subscribe to the action types created by `redux-thunk`.

I wanted to create this issue since it seems like it would clean a bit more the frontend code and remove duplication but is not super urgent to implement since our API calls are working. @dynamic11 what do you think about this?

PR #569, opened 2020-07-23
[Feat] Add visibility toggle in forms

### LINKED ISSUE #495: Add privacy buttons to every card on Edit Profile Page
When a user first creates their profile, there's no indication on whether the information they are entering will be kept private or public, unless they go through the privacy statements or FAQs beforehand. 

We should add the privacy buttons on the Edit profile page so that new users can set up their privacy settings as they create their profile.  This way, they can edit their privacy settings from the Edit page as well. 

**Additional context**
image

image

### LINKED ISSUE #570: In the profile, when you hover over the pencil it says "private" instead of edit
**Describe the bug**

In the profile, when you hover over the pencil it says "private" 

**Expected behavior**
It should say, "edit" everywhere the pencil is used 
 
**Screenshots**
image

**Additional context**

### LINKED ISSUE #571: Change mouse over wording for privacy buttons
The privacy buttons should be a little more discriptive to help people really understand what they mean in terms of visility of the data to others on the platform. 

I recommend we make the followings changes:

From: private,  To: Card only visible to you
From: public, To: Card visible to everyone
From: connections, To: Card visible to connections only

image

image

image

BenJeau:
#### Changes
- Extracted the visibility toggle component
- Added the toggle to every section for the forms (closes #495)
- Changes the hover for the edit pen button to "Edit" (fixes #570)
- Fixes hover of visibility toggles (fixes #571)
> Note: the OL toggle depends on its profile card being seperated (so it will seem like it won't work right now, but it is updating the correct value in **visibleCards**), until the PR from Trevor to seperate the card is created/merged
#### Screenshots
image
image
image
image
image

### DIFF
--- a/services/frontend-v3/src/components/basicInfo/BasicInfo.jsx
+++ b/services/frontend-v3/src/components/basicInfo/BasicInfo.jsx
@@ -1,5 +1,4 @@
 import React from "react";
-import { useHistory } from "react-router";
 import {
   LinkedinOutlined,
   GithubOutlined,
@@ -9,16 +8,20 @@ import {
 import BasicInfoView from "./BasicInfoView";
 import { ProfileInfoPropType } from "../../customPropTypes";
 
-const BasicInfo = ({ data }) => {
-  const history = useHistory();
+const styles = {
+  icon: {
+    marginRight: 5,
+  },
+};
 
+const BasicInfo = ({ data }) => {
   const getButtonLinks = () => {
     const { linkedin, github, gcconnex, email } = data;
     const buttonLinks = {
       email: {
         textId: "profile.email",
         url: `mailto:${email}`,
-        icon: <MailOutlined style={{ marginRight: 5 }} />,
+        icon: <MailOutlined style={styles.icon} />,
       },
     };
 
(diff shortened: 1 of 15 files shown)


TrevorBivi:
I noticed that some times when you change the visibility setting of a card and refresh the page it isn't saved. Trying to see if there's an obvious reason I'm seeing this.

BenJeau:
@TrevorBivi In this branch specifically or in dev? Regardless it should be fixed (probably in this PR), just wanted to know if I introduced it

TrevorBivi:
@BenJeau It seem to only be seeing it on this branch and not dev.

TrevorBivi:
I figured out how to reproduce it. It seems that if you make multiple changes and then refresh only the last change is applied. Looking at the put requests being sent, the values being sent for visibility values only seem to show the change for the last card. The others are set back to the original value. This happens both on forms and when editing visibility straight on the profile.

BenJeau:
Oh that makes sense, I have an idea on how to fix it. Thanks for finding that!

BenJeau:
Now works as expected (but is modifying the params, which is not great, its currently done like that on dev), once we move to something like redux-thunk, this should be fixed, #448. @TrevorBivi can you retry to see if it works?

TrevorBivi:
It's working perfectly for me now

TrevorBivi (approved):

PR #572, opened 2020-07-23
[FIX] Fix experience field

### LINKED ISSUE #367: As an employee, the limit for describing my experience should be set at 250 words and make it non-mandatory
**Describe the bug**
The current limit is set at 250 characters. This is too low. 
As a user, I would like to add where I worked and save it. And then go back later to describe what I did in my job. 

To resolve this issue:
1. Change limit to 250 words for the description field
Edit english footnote " Max length 250 characters" to "Maximum length 250 words" 
Edit french foonote " Maximum de 250 caractères" to " Maximum de 250 mots"

2. make the Description field non-mandatory
Remove the red asterick next to Description title

**Screenshots**
image

TrevorBivi:
Increases max characters for an experience items description to 1500 from 250
Allows wordbreaking to prevent text for overflowing

wide description view

narrow description view


BenJeau (approved):
Looks good and works great!

PR #578, opened 2020-07-24
Feat mentor skills search

### LINKED ISSUE #290: Add toggle switch and field for finding mentors with specific mentorship skills in the results page
Users need to find mentors with specific mentorship skills. 

On the results page:
Add a toggle switch for searching mentors similar to EX-Feeder
Once user turns on mentor, there should be another field that appears for user to enter mentorship skills required. 

**Additional context**

image

image

TrevorBivi:
Adds ability to search for specific or any mentor skills

_Adds mentor skills search fields where second temp classifications field was_
mentorSkillsSearch

_Adds mentor skills search fields to filters and reorders filters to better closer match order in advanced search_
filterResultsPage

_Skills now uses a tree select (mentor skills does too)_
TreeSelectForSkills


services/backend/src/core/search/util/anyMentorSkillsSearch.js
```diff
@@ -0,0 +1,13 @@
+async function anyMentorshipSkillSearch(profiles) {
+  const skillProf = [];
+
+  profiles.forEach((profile) => {
+    if (profile.mentorshipSkills && profile.mentorshipSkills.length) {
```
BenJeau: I think array.length is only thruthy when its equal to 1
```suggestion
    if (profile.mentorshipSkills && profile.mentorshipSkills.length > 0) {
```

services/frontend-v3/src/i18n/fr_CA.json
```diff
@@ -179,6 +180,8 @@
 
   "more.content.soon": "Plus de contenu à venir bientôt!",
 
+  "select.any": "Toute",
```
BenJeau: There's not really a good translation for any in french

```suggestion
  "select.any": "Ayant des compétences de mentorat",
```

services/backend/src/core/search/search.js
```diff
@@ -72,6 +74,14 @@ async function filterSearch(request, response) {
       results = await utils.classificationSearch(results, classifications);
     }
 
+    if (!anyMentorSkills && mentorSkills) {
+      results = await utils.mentorSkillsSearch(results, mentorSkills);
+    }
+
+    if (anyMentorSkills) {
+      results = await utils.anyMentorSkillsSearch(results);
+    }
```
BenJeau: Should treat anyMentorSkills like a string
```suggestion
    if (anyMentorSkills === "true") {
      results = await utils.anyMentorSkillsSearch(results);
    } else if (mentorSkills) {
      results = await utils.mentorSkillsSearch(results, mentorSkills);
    }
```

BenJeau (commented):
Small bug, while search with any, then removing it, look at the last comment, should probably fix it. Aside from that looks good and works!

services/backend/src/core/search/util/anyMentorSkillsSearch.js
TrevorBivi: All numbers greater than 0 are truthy, but that does look more clear

BenJeau (approved):

PR #603, opened 2020-07-28
[Fix] Routes update

### LINKED ISSUE #248: navbar avatar visible even when logged out
**Describe the bug**
the navbar avatar is visible even when logged out and on landing page. It does force the user to login again if they try to use the dropdown but it shouldn't be there in the first place.

**To Reproduce**
I believe this only starts occurring when keycloak session times out but the application thinks you are still logged in

**Expected behavior**
If you are not logged in it should say "sign in"  instead of avatar

**Screenshots**
image

image

**Additional context**

### LINKED ISSUE #521: Clean up frontend routes
Right now we have very long routes that are going to get messy as the app grows. I propose we change to the following routes:

domain.com => home 
domain.com/login =>login
domain.com/profile/:id =>profile 
domain.com/admin => admin 

I think we should try to keep the domains as short as possible since users might want to share them with others having a very long domain might be difficult. If anyone has any opinions on this please feel free to comment. 

relates to: #248

mo-5:
- Simplified routes. 
- Removed routes logic from app.jsx. 
- Reuse keycloak token (renew when expired instead of requesting a new token)
- Simplified keycloak logic in frontend 
- No longer able to go to the landing page once the user is logged on.

This closes #521, closes #248.


services/frontend-v3/src/App.jsx
```diff
@@ -75,33 +70,23 @@ const App = () => {
...
+    initConfig={keycloakConfig}
+    LoadingComponent={() => <AppLayout loading />}
+  >
+    <ReduxProvider store={store}>
+      <PersistGate loading={null} persistor={persistor}>
+        <App />
+      </PersistGate>
+    </ReduxProvider>
+  </KeycloakProvider>
 );
```
BenJeau: Could probably add the `<IntlProvider />` to this, and should probably rename the const to Provider or/and just create a new component that just wraps its children in another file (the second option is probably the cleanest. it would probably debloat the App.jsx

BenJeau (commented):
Everything looks good, just the single comment I left

BenJeau (approved):

BenJeau:
There's just a lot of exhaustive-deps warning when compiling/in the console logs, maybe you should fix that (maybe just add an eslint disable?)

mo-5:
> There's just a lot of exhaustive-deps warning when compiling/in the console logs, maybe you should fix that (maybe just add an eslint disable?)

I didn't notice them. I will take a look. I am not sure if they caused by my changes but I will look into it

PR #607, opened 2020-07-29
[Fix] Search undefined behavior

### LINKED ISSUE #606: Can search with empty text field, resulting in undefined string
**Describe the bug**
When searching in the general search, when viewing a profile, you can "search" with an empty field, it will just add "undefined" as string.
**To Reproduce**
1. Go to any profile
2. Search in the general search in the middle top without typing anything in the search box
3. See "undefined" be filled up

**Expected behavior**
To not be able to search with empty text field

BenJeau:
fixes #606


KateAThornley (approved):
Nice catch!

PR #628, opened 2020-07-30
[Fix] Add regex rules to firstName and lastName in form

### LINKED ISSUE #617: Users are able to enter any character for first and last name
**Describe the bug**
1. user can provide any first and last name.  You are able to change your name to anything including a space or numbers

**To Reproduce**
- Navigate to My Profile -> General Profile Info -> Edit Profile

**Expected behavior**
We should set some rules for first and last name. 

**Describe the bug**

2. Note that your profile name and picture does not update with these changes

**To Reproduce**
- Navigate to My Profile -> General Profile Info -> Edit Profile
- Change Name
- Save and Finish

image

BenJeau:
### Changes
- Added regex for firstName and lastName to accent only valid characters, and if a "-" is present, it needs to be surrounded by letters
- Capitalize firstName and lastName in the backend before saving to the database
- Get name from database to display in the dropdown

fixes #617

### DIFF
--- a/services/backend/src/core/profile/profile.js
+++ b/services/backend/src/core/profile/profile.js
@@ -27,10 +27,6 @@ function idHelper(id, savedId) {
   };
 }
 
-function capitalizeFirstLetter(string) {
-  return string.charAt(0).toUpperCase() + string.slice(1);
-}
-
 async function updateProfile(request, response) {
   try {
     validationResult(request).throw();
@@ -208,8 +204,8 @@ async function updateProfile(request, response) {
       await prisma.user.update({
         where: { id: userId },
         data: {
-          firstName: firstName ? capitalizeFirstLetter(firstName) : undefined,
-          lastName: lastName ? capitalizeFirstLetter(lastName) : undefined,
+          firstName: firstName ? _.upperFirst(firstName) : undefined,
+          lastName: lastName ? _.upperFirst(lastName) : undefined,
           teams: teams
             ? {
                 set: teams,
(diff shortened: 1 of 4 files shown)


services/backend/src/core/profile/profile.js
```diff
@@ -27,6 +27,10 @@ function idHelper(id, savedId) {
   };
 }
 
+function capitalizeFirstLetter(string) {
+  return string.charAt(0).toUpperCase() + string.slice(1);
```
mo-5: Should we just use lodash here since its already imported? _.upperFirst

services/backend/src/core/user/user.js
```diff
@@ -67,6 +67,10 @@ async function getUserById(request, response) {
   }
 }
 
+function capitalizeFirstLetter(string) {
+  return string.charAt(0).toUpperCase() + string.slice(1);
```
mo-5: Same here

services/backend/src/core/profile/profile.js
```diff
@@ -204,8 +208,8 @@ async function updateProfile(request, response) {
       await prisma.user.update({
         where: { id: userId },
         data: {
-          firstName,
-          lastName,
+          firstName: firstName ? capitalizeFirstLetter(firstName) : undefined,
```
mo-5: If we set it to undefined wouldnt that erase the name from the db ?

BenJeau: No, it would just act as there would be no variable there. If its null, it will make it null in the db

services/backend/src/core/profile/profile.js
BenJeau: Ohh didn't know there was a lodash function for that... will make the change

mo-5 (approved):

Issue #524, opened 2020-07-10
Add position title field to Profile 


MenaMachado:
**Is your feature request related to a problem? Please describe.**

Users can manually add their job title to their profile

**Describe the solution you'd like**

- On the Edit profile page in Contact information card, add a field under the first name to manually add their position title.
- In the user profile the title should be displayed just below the user's name

![image](https://user-images.githubusercontent.com/49656364/87206956-ba260600-c2d8-11ea-8e47-e36759075ac6.png)

![image](https://user-images.githubusercontent.com/49656364/87206906-8fd44880-c2d8-11ea-9b2c-a54de8ce037e.png)

MenaMachado:
Upon save pop up modal to offer 3 options:
leave second language as is
enter custom title for second language
Single language for both french and english

PR #640, opened 2020-07-31
[Feat] Made employmentInfo updatable with a single lang

BenJeau:
This is for @KateAThornley


BenJeau:
related to #524

KateAThornley (approved):
Thanks!

PR #649, opened 2020-08-04
[Fix] docker fronted crashing

### LINKED ISSUE #650: Docker frontend container crashing
**Describe the bug**
the fronted docker container crashes. the logs show that the node packages cannot be found.

dynamic11:
- docker fronted was crashing due to node_modules not being found in the container 
- issue was related to volume naming
- reorganized docker file commands to improve performance
- cleaned up docker files
- add docker ignore files
- closes #650


services/frontend-v3/Dockerfile
```diff
@@ -2,15 +2,21 @@ FROM node:12-alpine
...
 
-# Copies code into container
+# Set working directory
 RUN mkdir /upskill-frontend
 WORKDIR /upskill-frontend
-COPY . .
+
+# Copy app dependencies and scripts
+COPY package*.json ./
+COPY ./scripts ./scripts/
```
BenJeau: I'm curious as to why you are not copying the whole directory first

BenJeau (commented):
I'll test it locally, and also try without copying specific things in the Dockerfile (will remove node_modules on my machine to properly test it)

services/backend/Dockerfile
```diff
@@ -2,11 +2,15 @@ FROM node:12-alpine
...
 
-# Copies code into container
+# Set working directory
 RUN mkdir /upskill-backend
 WORKDIR /upskill-backend
+
+# Copy app
 COPY . .
 
+COPY ./scripts ./scripts/
```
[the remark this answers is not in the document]
dynamic11: ooooops

services/frontend-v3/Dockerfile
dynamic11: i looked at bunch of documentation and they consistently do this. I tried looking around for a reason but i cant find any  explanation other than possible performance improvements. 

Here is one of many examples and a possible explanation ( I should add dockerignore file too):
image

Let me know what you think

dynamic11: I also found this:
image

BenJeau: Just saw that exact article haha, keep it as is, but can you do the same for the backend dockerfile then?

dynamic11: thats a bit more difficult since you are using yarn generate which needs the database directory. Its 100% doable but it I will have to move stuff around a bit.

BenJeau:
From the last article that you just refferenced, it says to specify an absolute version of a FROM dockerfile, could you specify the latest lts one, which is `node:12.18.3-alpine3.11` for the backend and the frontend

services/frontend-v3/Dockerfile
BenJeau: mhhh you're right, then maybe we could just copy the src/database/ also

services/backend/Dockerfile
```diff
@@ -1,18 +1,26 @@
...
 # Set working directory
 RUN mkdir /upskill-backend
 WORKDIR /upskill-backend
 
-# Copy app
-COPY . .
+# Copy app dependencies
+COPY package*.json ./
+
+# Copy database directory for 'yarn geenrate'
```
BenJeau: Small typo

```suggestion
# Copy database directory for 'yarn generate'
```

BenJeau:
Rebuilding it (taking a while on my super fast internet :stuck_out_tongue:)

BenJeau (approved):
Seems to work!

PR #664, opened 2020-08-05
[Fix] seperate install and build script

### LINKED ISSUE #663: Build command fails in openshift
**Describe the bug**
Build command fails in openshift because the project is not copied. To fix this we have separate the build and install commands into separate scripts.

dynamic11:
- this should resolve the issue where openshift could not build the image
- closes #663


BenJeau (approved):
Works on openshift

Issue #563, opened 2020-07-22
Start the process to converting backend from Javascript to Typescript


mo-5:
Start the setup process to convert the backend from Javascript to Typescript. Typescript will offer better support to the project with type checking and better error handling. This would be the best way to catch bugs before deploying code.

mo-5:
This is a very low Priority I think it would just be good to implement. I just opened the ticket to not forget about it.

BenJeau:
I'm starting to convert the backend to typescript

PR #665, opened 2020-08-05
[Feat] Migrate part of backend to typescript

BenJeau:
related to #563 

Only part of the backend has been migrated to typescript to start using its benefits and in hopes to not cause major conflicts with others playing with the backend


mo-5 (approved):

BenJeau:
Waiting on @dynamic11 to talk to the maintenance team. Please put any news about it in this PR (or the related issue) so we log it / for others to know. Also, we should make a decision as fast as possible, since the conflicts will become big

PR #674, opened 2020-08-06
[Fix] Removed unused api/stats/hiddenUsers

BenJeau:
This API seems to not be used anywhere in the application (even though it is being called, its data is not being used)


mo-5:
Shouldn't this API be used here for the hidden user's section? 

image

BenJeau:
You made me think twice haha, this is the `api/stats/hiddenUsers` which returns a list of hidden users, the API that is called in the frontend for the stats cards is `api/stats/count/hiddenUsers`. No where in the app we are showing just the hidden users (and don't really see the use of it)

mo-5 (approved):

PR #691, opened 2020-08-07
[Fix] Page height

### LINKED ISSUE #685: Appearance of blank screen when user completes create profile steps
The Qualification section of the user edit profile page is long. When a user is done creating the profile, they are at the bottom of the page. 

Below is what a user sees when they click on the "save and finish" button of the qualification section. In user testing, users were not sure what to do at this point.  However, if you scroll up you will see the "done" message (see below). 

**For solution**

When user clicks on "save and finish" we should bring user back to top of page. 

Can we automatically bring user up to the top of page
image

Done message at the top of page. 
image

As you can see the Qualification form is long and the users is at the bottom of the page when done. 
image

BenJeau:
### Changes
- Scroll to top everytime the pathname in the url changes
- Fixed bug where the footer is somehow not visible (because of the marginTop: 64 in the Affix of the SideNav - this is only needed in the admin view for some weird reason)

fixes #685


mo-5:
Do you think we should pass isAdmin around the application or use @react-keycloak/web to check their type of authentication?

BenJeau:
I should maybe clarify the use of that prop, or rename it. Its only true if its rendering an admin view (since there's a bug for some reason with the admin side nav)... I think I'll remove the prop and just check the pathname if it contains "admin".

mo-5 (approved):

PR #709, opened 2020-08-07
[Fix] Properly hide hidden profiles

### LINKED ISSUE #669: Hidden/inactive profiles should not be visible to other (via profile link)
**Describe the bug**
A hidden/inactive user is still accessible via their profile link

**To Reproduce**
Kind of hard to reproduce with yourself only, but if someone is hidden/inactive, you can see their profile with their link

**Expected behavior**
To "act" as if the profile doesn't exist

BenJeau:
### Changes
- Return a 404 from the backend if that profile is hidden (and the user requesting to see that profile is not an admin)
- Properly handle case where the profile does not exist (instead of redirecting the user to their profile, show this)
image
- Make every error in the `<AppLayout />` so it doesn't confuse the user
ex: forbidden before
image
after
image

fixes #669


mo-5:
I like these changes. I know you did write this but can we change "Looks like someone doesn't belong" to "Your account doesn't have the correct permissions to view the following page" Or something formal.

services/frontend-v3/src/components/errorResult/errorResult.jsx
```diff
@@ -36,16 +40,43 @@ const ErrorResult = ({ errorCode }) => {
         </Button>
       ),
     },
+    profileNotExist: {
+      status: "404",
+      title: <FormattedMessage id="profile.not.found" />,
+      subTitle: <FormattedMessage id="profile.not.found.description" />,
+      extra: (
```
mo-5: Do you think we should break this into different components? One for each error page and then errorResult.jsx could use the different components?  I think having each component have one task would be ideal.

BenJeau: So lets say we keep the `<ErrorResultView />` and have something like `<404 />`, `<403 />`, and `<ProfileNotFound />`?

mo-5: Yeah, this would give us scalability as well since if we need to add more we could always create more components

mo-5 (approved):

PR #711, opened 2020-08-07
[Fix] Fixed broken link on Talent Management Form tooltip

### LINKED ISSUE #680: The link to the Talent Management tool appears to be broken in the Edit profile page
**Describe the bug**
In the Talent mangement results section, there is a tool tip with a link to the Talent Management tool. That links seems to be broken. 

The tool is a PDF form saved on ISED's internal website. 

**Solution:**
- Instead of using the link to the form, use this link to ISED's website : http://icweb.ic.gc.ca/eic/site/078.nsf/eng/h_00075.html   and http://icweb.ic.gc.ca/eic/site/078.nsf/fra/h_00075.html for french.
- Remove the following text at the end of the tool tip sentence: "following form"

**Screenshots**
image

KateAThornley:
- Changed out Link tags for a tags for Talent Management Form tooltip


BenJeau (commented):
Ohh good catch, could you make it so it opens a new tab with `target="_blank"` so the user doesn't lose their progress if they didn't save (since the a tags won't catch that)?

BenJeau (approved):

PR #720, opened 2020-08-09
[New] Display validation errors in notification in profile forms

### LINKED ISSUE #615: Increase size of changes saved success message
When you save changes in edit profile, you get a success message.  This message appears close to the search field and is a little small. 

I think we should increase the size of the message and perhaps have it slide in from the right above employee summary card where it is more visible. 

**Additional context**
image

### LINKED ISSUE #719: Form validation not visible for tabbed forms
now that "edit profile" forms are tabbed if there are validation errors in hidden forms then user get error but they dont know where the problem is.  

- change the error "message" to "notification" component
- display all validation errors in notification
- highlight tab that has validation error

[antd notification component](https://ant.design/components/notification/#header)

dynamic11:
- replace "message" component with "notification" component form top right of page
- Print name of forms with issues in the notification
- form tabs will highlight red if there is an error to get user attention
- closes #719, closes #615 

image
image
image


BenJeau:
I like the new notifications prompts, but they don't seem entirely translated

ex: 
image

---------
Also, idk if its just me, but the education 1 in the form item titles is a bit redundant (same here, the rules message and the form titles were translated in the past)
image

-------------
Not exactly sure how you implemented this, but when you address the problems (by either deleting what you did or by fixing it), the tabs are still red
image

---------
This is probably not you, but could you also translate the 'Mentorship Skills' form item?
image

--------

Since you are introducing this new prompt/notifications, could we entirely remove the old messages? This would fix #615

ex (when you save and finish in the edit forms, all the message popups in the admin dash, ..): 
image

BenJeau:
Upon looking at the code, you are inputting variables into the rule messages, which is something that can be done with `react-intl`, see [this](https://formatjs.io/docs/react-intl/components#formattedmessage) and you can also see how I've used it in `UpSkill/services/frontend-v3/src/components/profileForms/primaryInfoForm/PrimaryInfoFormView.jsx:616`

dynamic11:
- I will implement the translation thing today. Thank you for the info.
- the reason i have to say "education 1" in the validation error is that the same message is showed in the notification and the person should know which item its coming from. I can do some tricks but i decided it doesn't hurt and will keep the code change simple. I think we should merge this change and improve on it later if we decide its really bad.
- about the tabs staying red: let me look into it. I would have to revalidate "on change". is that something we want?
- about the new notification: yes, we should migrate the rest of the app to the new notifications. I think its out of scope of this PR but we/I can open a new PR to change it everywhere else. lets first see if they like this notification or they want to open it at bottom of the page.

BenJeau:
Small visual bug, it seems like there's extra bottom padding when displaying errors in the notification

dynamic11:
@BenJeau are your "edit profile" forms laggy as you type? mine seems to be laggy but I'm not sure if its my computer or the app.

BenJeau:
yes, acting form section is slow for some reason and the qualification forms - 
image

And when I saved that form, I got this error -
image

BenJeau:
Development is not laggy - so I guess its this branch - its probably re-rendering a lot of times, I'll quickly look at your code

services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
```diff
@@ -217,21 +227,28 @@ const EmploymentDataFormView = ({
...
   };
 
-  /* show message */
-  const openNotificationWithIcon = (type) => {
+  /*
+   * Open Notification
+   *
+   * open notification to show status to user
+   */
+  const openNotificationWithIcon = ({ type, description }) => {
```
BenJeau: I think these types of comments hinder code readability (and maintability - since people need to update them). As long as the function/variable name explains well what is does/contains, we should probably not put comments for them. What do you think about that?

Here (since it doesn't look like we're gonna be moving to typescript - at least for the frontend), putting JSdocs for the params/return instead would be more useful. (Little side note, for comments to be seen when using a function/variable, they need to start with /**, to create a JSdoc comment).

Also, I'm curious as to why you passing an object to this function instead of two distinct param.

----
Ex (if it would be two params):
```suggestion
  /**
    * @param {"success" | "error" | "warning"} type Notification type
    * @param {string} description Text to be displayed in the notification
    */
  const openNotificationWithIcon = (type, description) => {
```

which would render like this in vscode when using/hovering the function
image
image


instead of 
image

services/frontend-v3/src/components/profileForms/personalGrowthForm/PersonalGrowthFormView.jsx
```diff
@@ -157,6 +158,14 @@ const PersonalGrowthFormView = ({
...
   };
 
+  /* Component Rules for form fields (commented out as example) */
+  // const Rules = {
+  //   required: {
+  //     required: true,
+  //     message: <FormattedMessage id="profile.rules.required" />,
+  //   },
+  // };
+
```
BenJeau: Should probably remove this

services/frontend-v3/src/components/profileForms/personalGrowthForm/PersonalGrowthFormView.jsx
```diff
@@ -242,21 +257,88 @@ const PersonalGrowthFormView = ({
...
+    // errors.forEach((value) => {
+    //   errorArray["learningDevelopment"] =
+    //     errorArray["learningDevelopment"] ||
+    //     (String(value.name).includes("learning") && value.errors.length > 0);
+    // });
+
+    const errorArray = [];
+    // save results to state
+    setTabErrorsBool(errorArray);
+    return errorArray;
```
BenJeau: Is this useful, seems like you are doing it below?

services/frontend-v3/src/components/profileForms/qualificationsForm/educationForm/EducationFormView.jsx
```diff
@@ -173,7 +172,7 @@ const EducationFormView = ({
         <Form.Item
           name={[field.name, "diplomaId"]}
           fieldKey={[field.fieldKey, "diplomaId"]}
-          label={<FormattedMessage id="profile.diploma" />}
+          label={`Education ${  field.name + 1  } Diploma`}
           style={style.formItem}
```
BenJeau: These fields should probably be translated

services/frontend-v3/src/i18n/en_CA.json
```diff
@@ -251,6 +252,7 @@
   "profile.rules.max.1500": "Max length 1500 characters",
   "profile.rules.max.50": "Max length 50 characters",
   "profile.rules.max.60": "Max length 60 characters",
+  "profile.rules.max": "Max length {max} characters",
```
BenJeau, line 255: I'm not sure if this is out of scope for this PR, if you introduce this, should we remove the other "profile.rules.max.XX"?

BenJeau (commented):
Just by looking at your code, I'm not sure why the forms are laggy, maybe rebasing it with dev will fix the lagginess 🤷

dynamic11:
> Development is not laggy - so I guess its this branch - its probably re-rendering a lot of times, I'll quickly look at your code

I think it has to do with the check I do for errors on change. but ill let you give you idea aswell before i move forward with a fix.

services/frontend-v3/src/i18n/en_CA.json
dynamic11: I want to remove it but let me make another PR for it so i can test it better and isolate it

services/frontend-v3/src/components/profileForms/personalGrowthForm/PersonalGrowthFormView.jsx
dynamic11: The reason I kept it in there was so it can be used as reference. Currently those forms dont have validation but if they would need it in the future then it must be uncommented. I was just trying to be "consistent" across all forms. I was on the fence about just removing it.

BenJeau: I personally don't like future proofing code, in case that feature will never get done - unless you know this is something we'll for sure need in the non-distant future, we should probably remove it. I think the other forms give a good enough example on how to enable rules

BenJeau: I would even go a step further, to remove code duplication, to have a single "Rules" object for all of our forms, instead of having to need to redefine it in every form

services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
dynamic11: on the topic of the comment. I was just trying to be consistent with the rest of the code. If we choose to change it  then we should do it for all functions in those files. I agree with  JSdocs for the params/return.

with regards to the object as the input param of the function, my reasoning was similar to what this guy is saying in this article. But I can change it. I dont mind.
https://medium.com/@david.gilbertson/named-function-arguments-61b9f88bd5e5

dynamic11:
> yes, acting form section is slow for some reason and the qualification forms -
> image
> 
> And when I saved that form, I got this error -
> image

Fixed!

services/frontend-v3/src/components/profileForms/personalGrowthForm/PersonalGrowthFormView.jsx
dynamic11: ^ will do this in another PR

services/frontend-v3/src/components/profileForms/qualificationsForm/QualificationsFormView.jsx
```diff
@@ -208,24 +208,19 @@ const QualificationsFormView = ({
...
+    const errorObject = form
+      .getFieldsError()
+      .reduce((acc, { name, errors }) => {
+        if (errors.length > 0) {
+          acc[name[0]] = true;
+        }
+
+        return acc;
+      }, {});
+
```
dynamic11: I really like this!

dynamic11:
ready for one more round of review @BenJeau !

BenJeau:
I would review it now, but keycloak seems down live :(

dynamic11:
Just connect to keycloak prod. I can send you the token if you like

BenJeau:
Its good, I have it

BenJeau:
I'm still not able to connect to keycloak, even with prod, I think its rejecting me since I'm making the request from localhost, instead of from openshift

dynamic11:
It works for me. I can send you my env files if you like. Also make sure you restart docker compose

services/frontend-v3/src/i18n/fr_CA.json
```diff
@@ -174,6 +174,7 @@
   "profile.diploma": "Diplôme",
   "profile.edit": "Modifier",
   "profile.edit.save.error": "Données non enregistrées",
+  "profile.edit.save.error.intro": "Les éléments suivants nécessitent votre attention:",
```
BenJeau:
```suggestion
  "profile.edit.save.error.intro": "Les formulaires suivants nécessitent votre attention:",
```

BenJeau (commented):
Just tested it locally, works great! (a bit laggy, but to be expected) And reviewed your code, looks great, just one small translation thing! fix that and I'll approve it

BenJeau (approved):

PR #729, opened 2020-08-10
[Fix] Cleaned up admin user dash

### LINKED ISSUE #724: Clean up admin user table (make less wide)
The admin dash user table is getting big, let's try to clean it up

- Remove the view profile buttons to make the names links
- Remove multiselect delete column and header button, to just create a button per row, like the view profile ones
- Shorten up the dates
- Add default role option, to filter them

BenJeau:
### Changes
- Removed multiselect delete column
- Remove view profile button (now need to click on the user name)
- Added "standard" user role - to display something when the user doesn't have the admin or manager role
- Added delete button for every row

before
image

after
image

fixes #724


services/frontend-v3/src/components/admin/userTable/UserTableView.jsx
```diff
@@ -121,17 +119,29 @@ const UserTableView = ({
...
+            autoEscape
+            textToHighlight={text.toString()}
+          />
+        ) : (
+          text
+        );
+
+      if (linkKey && record[linkKey]) {
+        return (
+          <Link to={record[linkKey]} target="_blank">
```
mo-5: I dont know if we should open the link in a new tab. The standard is if you are opening an external link it should be opened a new tab but if it is internal it should be on the same tab. The idea is the user would redirect to that user in the same window instead of having to switch between tabs. What do you think?

BenJeau: Makes sense to keep them inside the same tab, will remove the target

mo-5 (approved):

PR #739, opened 2020-08-10
[Feat] Add skip profile creation

### LINKED ISSUE #678: Add third option for new users to sign in but skip profile creation and go directly to search page

Users (Managers) will want to use I-Talent to search for employees for staffing positions. They may not have the time to create a profile and we want to encourage the use of the application. 

Currently, the app is designed to encourage profile creation and so they are taken through the steps to create a profile. Managers are forced to go through every step, by clicking save and next until done, leaving all sections empty until done.  Managers can find this cumbersome. 

**Solution**

On the welcome page, add a third option (button) for users (Managers) at first login to go directly to the search profiles. They would skip the profile creation page. 

Can we add the same "search profiles" button as when profile all done on the Welcome page?

image

image

BenJeau:
### Changes
- Created a very faint skip profile creation button
image
- Shows a popup to the user
image
- If the user selects yes, their profile will be hidden, and they won't need to fill up the forms
- Since "HIDDEN" is a user only option, admins will be able to see that option, but not set it in the table
image
- Updated wording according to related issue

fixes #678


mo-5:
I would say make it look clickable so the user is not confused by the button what do you think? You can still make it as a small button. Since right now it just looks like text but I am not sure if that's the objective.

BenJeau:
When you hover on it, you see its a button, but yeah I made it not look like a button on purpose for people to not be "attracted" towards it. @dynamic11 what do you think about this? Should it be more visible that its clickable?

services/frontend-v3/src/i18n/fr_CA.json
```diff
@@ -333,12 +333,14 @@
...
   "setup.welcome": "Bienvenue",
-  "setup.welcome.action": "Veuillez sélectionner un profil SAGE pour automatiquement remplir vos informations ou commencer de zéro",
+  "setup.welcome.action": "Veuillez sélectionner un profil GCannuaire pour automatiquement remplir votre organisation et vos coordonnonées ou vous pouvez commencer de zéro et synchroniser avec GCannuaire plus tard",
   "setup.welcome.description": "Nous avons besoin de quelques informations pour configurer votre profil",
-  "setup.welcome.geds.description": "de SAGE",
+  "setup.welcome.geds.description": "de GCannuaire",
   "setup.welcome.geds.title": "Récupération de profils",
   "setup.welcome.new.description": "débuter",
   "setup.welcome.new.title": "Nouveau utilisateur",
+  "setup.welcome.skip": "Sauter la création de profil ...",
```
mo-5: I would personally remove the '...' 
```suggestion
  "setup.welcome.skip": "Sauter la création de profil",
```

services/frontend-v3/src/i18n/en_CA.json
```diff
@@ -333,12 +333,14 @@
...
   "setup.welcome": "Welcome",
-  "setup.welcome.action": "Please select a GEDS profile to pre-populate your information or start from scratch",
+  "setup.welcome.action": "Please select GCdirectory profile to autofill your organization and contact information or you can start from scratch and sync with GCdirectory later",
   "setup.welcome.description": "We just need a few bits of information to set up your profile",
-  "setup.welcome.geds.description": "from GEDS",
+  "setup.welcome.geds.description": "from GCdirectory",
   "setup.welcome.geds.title": "Fetching profiles",
   "setup.welcome.new.description": "start fresh",
   "setup.welcome.new.title": "New user",
+  "setup.welcome.skip": "Skip profile creation ...",
```
mo-5: here as well
```suggestion
  "setup.welcome.skip": "Skip profile creation",
```

dynamic11:
I personally think it's good the way you implement it. I would remove the "..." though. I don't think other websites have the "..."

services/frontend-v3/src/components/profileForms/welcome/WelcomeView.jsx
```diff
@@ -68,6 +77,10 @@ const WelcomeView = ({ gedsProfiles, intl, load, userId, history }) => {
       fontStyle: "italic",
       marginTop: "-4px",
     },
+    skipButton: {
+      marginTop: 20,
```
mo-5: I know you didn't write this but do you think you could make a WelcomeView.scss file and move these there?  So that way we are making a slow migration

mo-5 (approved):

PR #748, opened 2020-08-11
[Fix] Language expire

### LINKED ISSUE #742: Can't unselect expired button on OL card and consequently issue with toggle switch.
**Describe the bug**
It seems the expired button is selected and when you try to remove the selection, it doesn't remove it when you save and finish.  If you entered a date and try to unselect it, the date saves but the expired button stays selected.

Also, I can't untoggle even when I delete the results. 

I'm getting a "sucessfully saved" message. 

Could this issue be similar to the one we were having with the job mobility card?

**Screenshots**

image

TrevorBivi:
- Fixes not being able to save unchecking "expired - unknown date" checkbox
- Fixed known expired dates turning into unknown expired dates

note: In the backend, normalizeDate returns undefined for a falsey date, not null. Null is required to set the date to be empty in the table. This is why I conditionally either call normalize date or use null for the date value

fixes #742


BenJeau:
I would accept this PR - I understand for the `normalizeDate()`, but I am curious to why you are sending `moment.unix(0)` from the frontend, where you could just send null instead and remove that null checking in the backend - feels like an extra cleaning step in the backend to send back the data to the frontend - but I may be wrong

TrevorBivi:
> I would accept this PR - I understand for the `normalizeDate()`, but I am curious to why you are sending `moment.unix(0)` from the frontend, where you could just send null instead and remove that null checking in the backend - feels like an extra cleaning step in the backend to send back the data to the frontend - but I may be wrong

The expiry date is null when the user hasn't provided any expirey information. The unix time 0 is used as a special case to specify the grade is expired but the date of expirey is not known. I know there was a discussion about instead adding an extra value for the table to specify an unknown but expired date. I think we wound up going with this because if something unexpected happened the table could be holding contradictory information (unknown expired flagged true but also storing an expirey date) ?

I made a branch off of this code that could be merged in that would instead send undefined and null for undefined and unknown expiry info instead of null and the unix 0 special case to hide it from the front end (https://github.com/CDH-Studio/UpSkill/compare/fix-language-expirey...hide-unix-0-from-frontend?expand=1). We could still also go the route of adding an extra table value, I don't mind

BenJeau (approved):

BenJeau:
Okk makes more sense. It's just that using `moment.unix(0)` is not clear for someone who didn't work on it. Could we add a new column in the SeconLangProf table, like a boolean, to symbolize if the checkbox? I'll create a new issue regarding this

PR #749, opened 2020-08-11
[fix] basic info card edit button position

TrevorBivi:
- Adds an option floatRight to the editButton class to give the class float:right styling
- expands width of basic info card row holding button and makes the button use the new floatRight prop

_image with new position of edit button (same amount of pixels from the right as other cards)_ 
repositionEdit


BenJeau (approved):
LGTM

mo-5:
Wouldnt we always have it float right? If we don't just apply css to the button element instead of passing the element through components this will just make it messy. What do you think about this? I feel like it would simplify it.

TrevorBivi:
> Wouldnt we always have it float right? If we don't just apply css to the button element instead of passing the element through components this will just make it messy. What do you think about this? I feel like it would simplify it.

Making the edit button always float right doesn't negatively effect other places it's used, so I'm fine with it. I switched to always using scss styling for EditButton instead

mo-5:
Looks good. Thank you.

PR #756, opened 2020-08-12
[feat] improve description form item

TrevorBivi:
- Allows description form item to accept multiple rules instead of only one
- Adds a new prop maxLength with a default value of 1000 instead of gathering max length from provided rule

Note: this has to due with changes requested by Ali so he should probably be the one to review it


dynamic11:
hold this until my changes go through on Monday. Also please look through the console and make sure this component isn't causing console errors. I saw a few in the 'edit profile' forms related to the description field 😄

services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
```diff
@@ -721,11 +721,12 @@ const EmploymentDataFormView = ({
               <DescriptionFormItem
                 name="description"
                 fieldKey="description"
-                rule={Rules.maxChar1000}
-                value={profileInfo.description}
+                rules={[Rules.maxChar1000]}
+                maxLength={1000}
```
dynamic11: The only rule i can imagine this field having is the manLength, isRequired, minLength, can we somehow make props for there an avoid passing in "rules". Eventually we are moving the rules to as separate file anyways.

dynamic11 (changes requested):

services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
TrevorBivi: The user can now pass the following props instead of rules:
 - maxLength
 - maxLengthMessage
 - minLength
 - minLengthMessage
 - isRequired
 - isRequiredMessage

services/frontend-v3/src/components/profileForms/descriptionFormItem/DescriptionFormItem.jsx
```diff
@@ -4,48 +4,81 @@ import PropTypes from "prop-types";
...
+  fieldKey: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.arrayOf(PropTypes.any),
+  ]),
+  minLength: PropTypes.number,
+  maxLength: PropTypes.number,
+  isRequired: PropTypes.bool,
+  minLengthMessage: PropTypes.element,
+  maxLengthMessage: PropTypes.element,
+  isRequiredMessage: PropTypes.element,
```
dynamic11: i would change this to **PropTypes.node** but im not 100% sure. node is more general than node btu i often see node.

dynamic11 (changes requested):

services/frontend-v3/src/components/profileForms/descriptionFormItem/DescriptionFormItem.jsx
TrevorBivi: changed element to node

BenJeau (approved):

BenJeau:
@dynamic11 for us to be able to merge this PR, you'll need to remove your requested changes review (I think)

dynamic11 (approved):

dynamic11:
> @dynamic11 for us to be able to merge this PR, you'll need to remove your requested changes review (I think)

@BenJeau Sorry! It should be good now. 😄

PR #758, opened 2020-08-12
[fix] fix private description card visibility

### LINKED ISSUE #746: About me section is not grayed out
**Describe the bug**
If a profile has the about me not visible to other (PRIVATE or CONNECTIONS), people will still see a white box instead of the grayed out one and will think the user left it blank

**To Reproduce**
1. Go to any of the seeds profile - locally
2. You will see their about me section even though it is PRIVATE

**Expected behavior**
The About me card to also be grayed out - to act like the other cards when it is hidden

**Screenshots**
image

TrevorBivi:
- Fixes description card not being set to non visible when non private
- Adds description visibility values to seed data (VISIBLE for john, PRIVATE for mary)

fixes #746


BenJeau (approved):
Works great! LGTM

PR #759, opened 2020-08-13
[fix] Clean some console error on profile page 

dynamic11:
- clean error due to connection status proptype
- clean error due to first language proptype


services/frontend-v3/src/components/officialLanguage/OfficialLanguageView.jsx
```diff
@@ -52,7 +52,7 @@ const OfficialLanguageView = ({ firstLanguageInfo, secondLanguageInfo }) => {
 };
 
 OfficialLanguageView.propTypes = {
-  firstLanguageInfo: PropTypes.isRequired,
+  firstLanguageInfo: PropTypes.array.isRequired,
```
BenJeau: You'll need to specify what is in the array, `PropTypes.arrayOf(....).isRequired`

dynamic11: on it

BenJeau:
I think you'll need to make that proptype either .isRequired(), or add a defaultProp to it

BenJeau (approved):

PR #760, opened 2020-08-13
[New] Move from lodash to lodash-es and change the import 

### LINKED ISSUE #633: Change Lodash to Lodash-es  Frontend
Lodash-es is a much smaller package than lodash. It has been exported to es modules. It can be found in the following link. 
https://www.npmjs.com/package/lodash-es

This can improve the performance and load time for backend and frontend. 

Size difference 1.4 MB (Lodash) and 629 kB (Lodash-es) 

Let me know what you guys think about this change if you see any issues with this change.

mo-5:
The best way to import lodash in react is by a component that we will use for example: 

import {map} from "lodash-es"; 

this will improve the speed of the application since we won't need to import the whole library for one method.

closes #633


BenJeau (approved):
Looks good, you'll just need to update two other imports

mo-5:
> Looks good, you'll just need to update two other imports

Which imports?

mo-5:
oh nevermind I see

mo-5:
I found that there are unclosed issues with lodash-es that break some functionality so I moved back to lodash but with the new way of importing by use methods

PR #761, opened 2020-08-13
[feat] Add N/A secondary language grade option

TrevorBivi:
Adds a N/A option to secondary language grades

na-form
na-profile


BenJeau:
I may be missing what N/A implies - but should a grade be able to have N/A and expired? Like if they don't have a grade for a second language proficiency, should they be able to set a date and/or set it to expired?

TrevorBivi:
> I may be missing what N/A implies - but should a grade be able to have N/A and expired? Like if they don't have a grade for a second language proficiency, should they be able to set a date and/or set it to expired?

Oops, you're right. totally overlooked that

PR #765, opened 2020-08-14
Official lang improvments

TrevorBivi:
- Adds a new boolean unknownExpiredDate to langProfItems to signify an unknown expired date instead of a special case for the Date column
 - Adds N/A grade option (and disables date selection for N/A grades)


services/backend/src/database/migrations/20200810152146-optional-interested-in-remote/schema.prisma
```diff
@@ -485,6 +486,7 @@ model User {
   telephone            String?
   cellphone            String?
   manager              String?
+  description          String?
```
BenJeau, line 489: Why is this file modified, this seem to be an old migration?

BenJeau (commented):
Looks good, one small comment - testing it locally right now

BenJeau:
Works great locally! Two small comments:

1. When something is N/A, could we ommit the (Unexpired)?
image

2. Should the default values be N/A?
image

services/backend/src/database/migrations/20200810152146-optional-interested-in-remote/schema.prisma
TrevorBivi: I think backend changes got merged incorrectly at some point. This schema was missing description even though it was added after the description card change. This was causing unexpected steps to be included in the migration

BenJeau: Ok sounds good, just hope it won't frig up the migrations on the server :crossed_fingers:

TrevorBivi:
> Works great locally! Two small comments:
> 
> 1. When something is N/A, could we ommit the (Unexpired)?
>    image
Good catch, fixed

> 2. Should the default values be N/A?
>    image

I don't know if it's a good idea to make the default values N/A since they would only see those form items if they checked the box saying they have have been graded. I don't think it's too common for people to be graded on some areas but not others.

BenJeau (approved):
LGTM

PR #780, opened 2020-08-15
[Feat] Converted to react-app-rewired to razzle

### LINKED ISSUE #140: Investigate Bizchart vs Antv G2
- we are currently using bizchart for data visualization
- antd has a sister project for visualization called antv G2
- investigate if switching over is necessary/needed

**This is very low priority**

### LINKED ISSUE #779: Fix jenkins - Be able to have runtime environment variables for the frontend
Create react app and react app rewired doesn't allow for runtime environment variables, which is needed to only build the image once and use that same image thoughout all of the server envionments.

Here are some alternatives to react app rewired that allow runtime variables
https://razzlejs.org/docs/environment-variables#runtime-variables

BenJeau:
### Changes
- Added `razzle.js` and remove `react-app-rewired`, `react-scripts`, and testing libs (since we are not using them)
  - Razzle introduces SSR (server side rendering), the application is now served by expressjs
  - there are many checks to see if we are doing SSR, ex this is a check to see you doing SSR `typeof window === "undefined"` or `typeof document === "undefined"`
  - Razzle adds the ability to have runtime variables
  - Added custom webpack configs to overwrite less variables for ant design
- Updated frontend envs to be shorter (since they don't need to start with `REACT_APP_`)
- Removed unused styles in `App.scss`
- Updated graphing lib to latest version of Bizcharts
- Fix routing in the DoneSetupView form, to properly use `react-router-dom`
- Removed service worker
- Created frontend file containing config (mapping environment variables)
- Updated frontend packages to the most recent versions (except for `mini-css-extract-plugin` which needs to be @0.9.0 for razzle)
- Changed `@react-keycloak/web` to `@react-keycloak/razzle` (https://github.com/panz3r/react-keycloak/blob/master/packages/razzle/README.md) 

> __Note:__ Auto-reloading works like before, but since it is doing SSR, you may need to restart the docker container if you made some big changes

fixes #779, fixes #140


BenJeau:
Everything works as expected - may be a little more slow (but I think its just my laptop - edit: it was just my laptop). I just can't seem to be able to override ant design's theme

image

And our library for the graphs doesn't seem to support SSR

BenJeau:
@dynamic11 or @MohamedRadwan if you test this locally, you'll need to rebuild the frontend docker image for it to properly work and please update your envs in `web.env` to the same keys as in `web.env.example`. Also @dynamic11, I can't seem to be able to see the issue with snyk

services/frontend-v3/src/components/admin/dashboardGraphs/DashboardGraphsView.jsx
```diff
@@ -1,10 +1,19 @@
...
-import { Chart, Geom, Axis, Tooltip, Coord, Legend } from "bizcharts";
 import { FormattedMessage, injectIntl } from "react-intl";
 import { IntlPropType } from "../../../utils/customPropTypes";
 
+const { Chart, Geom, Axis, Tooltip, Coord, Legend, registerTheme } =
+  typeof document === "undefined" ? {} : require("bizcharts");
+
+if (registerTheme) {
+  //   registerTheme("default", {
+  //     defaultColor: "#505050",
```
mo-5: Is this commented code needed ?

BenJeau: Oups, I'll remove it - It was to change the colors for the charts

BenJeau:
I'll need to quickly cleanup App.js to look like their examples (I think), to do SSR correctly https://github.com/react-keycloak/react-keycloak-examples/tree/master/examples/razzle-app/src

BenJeau:
Branch/PR should be good to be tested

mo-5 (approved):

PR #783, opened 2020-08-17
[Fix] Form control button for last create form

BenJeau:
Forgot one scenario for the `<FormControlButtons />`, right now it would display "Save and next" for the qualifications form in the create profile flow, when it should be "Save and finish" and right now if you click on "Save and next" on that form, it would crash the app


dynamic11 (approved):
looks good!

PR #786, opened 2020-08-17
[New] Add/Edit/Delete Links to supporting documents under Education and Experience sections

### LINKED ISSUE #435: Add/Edit/Delete Links to supporting documents under Education and Experience sections
Employees will want to add links to documents such copies of certificates, resumes, projects material, etc.  Employees will also want to provide links to specific posters in jobs.gc.ca to show the jobs they qualified in.  They may also want to provide links to git repositories, etc. 

Rather than upload documents that take space, it is better for employees to save their documents in GCDOCs and provide links to those documents in I-Talent. 

Instructions

**For Education**

- For each education (e.g Ed 1) Users will want to add/edit/remove links to the following documents: diploma, transcript, publication list
- The display name for the URL can be: diploma, transcript, publication list
- There should be only one diploma,one transcript, and one publication list per Education
- Suggestion is to add a button for add link similiar to "add item"

**For Experience**
- For each experience (e.g EX 1 ) Users will want to add/edit/remove links to the following documents: Awards and recognition, Reference letter, Work Sample. 

The display name can be Work sample 1, Work sample 2, Work sample 3, etc. 

They may also want to add links to sites like Github, Bitbucket, Wiki, etc. 

image
image

Below are screens shots from GC Directory as an example for design, requirements and wording.

Edit/Delete Links

image

image

image

image

image

mo-5:
- Cleaned up qualifications form. 
- Used async api call to improve speed 
- Refactored the view and the forms 
- Moved to scss
- Set the maximum number of experience, links and eductions to 3 (can be modified later)
close #435

image
image
image


services/backend/src/database/schema.prisma
```diff
@@ -521,7 +568,7 @@ model User {
   educations           Education[]
   experiences          Experience[]
   relocationLocations  RelocationLocation[]
-  connections          User[]
+  connections          User[]                @relation("UserToUser")
   user                 User?                 @relation("UserToUser", fields: [userId])
```
BenJeau, line 574: Can you have relation names be identical as another one?

services/frontend-v3/src/components/education/EducationView.jsx
```diff
@@ -17,6 +17,19 @@ const EducationView = ({ educationInfo }) => {
...
 
+  const getUrl = (item) => {
+    if (item.attachmentLinks && item.attachmentLinks.length > 0)
+      return item.attachmentLinks.map((i) => (
+        <Tag color="rgb(114, 114, 114)" key={i.id}>
+          <LinkOutlined />
+          <a target="_blank" rel="noreferrer" href={i.url}>
+            {i.name}
+          </a>
+        </Tag>
```
BenJeau: Here I would make the tag clickable and I figured out to add proper spacing between the icon and the text, you need to wrap the text with a `<span></span>`

Maybe something like this?
```suggestion
        <a target="_blank" rel="noreferrer" href={i.url}>
          <Tag color="rgb(114, 114, 114)" key={i.id}>
            <LinkOutlined />
            <span>
              {i.name}
            </span>
          </Tag>
        </a>
```

services/frontend-v3/src/components/education/EducationView.jsx
```diff
@@ -17,6 +17,19 @@ const EducationView = ({ educationInfo }) => {
...
+      ));
+    return <></>;
```
BenJeau: Here you can just return null or undefined (to keep consistency with the rest of the code)

services/frontend-v3/src/components/experience/ExperienceView.jsx
```diff
@@ -11,6 +11,32 @@ const ExperienceView = ({ experienceInfo }) => {
...
+    if (item.attachmentLinks && item.attachmentLinks.length > 0)
+      return item.attachmentLinks.map((i) => (
+        <Tag color="rgb(114, 114, 114)" key={i.id}>
+          <LinkOutlined />
+          <a target="_blank" rel="noreferrer" href={i.url}>
+            {i.name}
+          </a>
+        </Tag>
+      ));
+    return <></>;
```
BenJeau: Same comment as above

services/backend/src/router/options/options.js
```diff
@@ -247,4 +248,11 @@ optionsRouter.get(
   tenures.getTenures
 );
 
+optionsRouter.get(
+  "/attachmentNames",
+  keycloak.protect(),
+  langValidator,
+  linkAttachmentNames.getNames
+);
```
BenJeau: Since you are getting a `type` param, you should create a new validator for it

services/frontend-v3/src/components/profileForms/qualificationsForm/linkAttachment/LinkAttachmentView.jsx
```diff
@@ -0,0 +1,80 @@
...
+        </Form.Item>
+      </Col>
+      <Col className="gutter-row" span={18}>
+        <Form.Item
+          name={[fieldElement.name, "url"]}
+          fieldKey={[fieldElement.fieldKey, "url"]}
+          className="formItem"
+          rules={[Rules.required]}
+        >
+          <Input placeholder="Attachment URL - ex. www.google.com" />
```
BenJeau: This placeholder should be translated

services/backend/src/core/profile/util/updateProfile.js
```diff
@@ -148,14 +148,93 @@ async function updateProfile(request, userId, language) {
...
+    const attachmentLinkId = flatten(
+      await Promise.all(
+        experiences.map((ed) =>
+          prisma.attachmentLink.findMany({
+            where: { experienceId: ed.id },
+            select: { id: true },
+          })
+        )
+      )
+    );
```
BenJeau, line 163: I'm curious, what is the use of `flatten`?

BenJeau (commented):
Looks good, just a few comments and testing it locally right now. And can you add to the swagger docs for the new API endpoint?

services/backend/src/core/profile/util/updateProfile.js
mo-5: The structure of this method was kind of confusing. So the return of this method was a nested array of depth 1 which made it hard to check for the length.

services/backend/src/database/schema.prisma
mo-5: what do you mean? for the UserToUser relation?

BenJeau: Yes, nevermind I think I understand, but why is this being added in this PR, is it just the vscode formatter that added this?

mo-5: Yeah, prettier automatically added it. Not sure why now maybe after an update of one of the packages.

mo-5:
Everything should be good. Ready for another review.

services/frontend-v3/src/components/experience/ExperienceView.jsx
```diff
@@ -11,6 +11,32 @@ const ExperienceView = ({ experienceInfo }) => {
...
   };
+  const getUrl = (item) => {
+    if (item.attachmentLinks && item.attachmentLinks.length > 0)
+      return item.attachmentLinks.map((i) => (
+        <Tag color="rgb(114, 114, 114)" key={i.id}>
+          <LinkOutlined />
+          <a target="_blank" rel="noreferrer" href={i.url}>
+            {i.name}
+          </a>
+        </Tag>
```
BenJeau: Can you put the a tag outside and add a surround the text with a span - for it to be like the education ones. For it to have a pointer cursor, you'll probably need to add `style={{cursor: 'pointer'}}` to the tag, and same thing for the education link

BenJeau (commented):
Works well! 

I think you didn't define `TypeQuery` for the docs, this is what is shown
image

And I think you need to update this branch with dev

mo-5:
Sounds good, should be all fixed up.

BenJeau (approved):

PR #795, opened 2020-08-19
[Feat] Added basic API request logger

BenJeau:
#### Changes introduced

Added basic API request logger - useful when debugging and checking what the frontend calls in the backend

#### Screenshots (if aplicable)

image

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible

### DIFF
--- a/services/backend/package.json
+++ b/services/backend/package.json
@@ -35,6 +35,7 @@
     "keycloak-connect": "^11.0.0",
     "lodash": "^4.17.20",
     "moment": "^2.27.0",
+    "morgan": "^1.10.0",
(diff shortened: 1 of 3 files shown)


mo-5:
Just wondering are we console logging all the requests ? Should we create a db table and push all these to it so we have it saved ?

mo-5 (approved):

BenJeau:
Yeah, its console logging the requests, but we can save it to a file instead and add more info, like ip address, and time for when its running on the server

PR #837, opened 2020-09-09
[Fix] Url field lag

### LINKED ISSUE #821: The URL attachment field input is not very responsive
**Describe the bug**
When inputting text in the URL field input, you can see that the webpage is not responding fast - I think this is something related to the `findErrorTabs()` function, since it is not doing a deep equals check.

**To Reproduce**
1. Go to 'Edit profile'
2. Click on 'Qualifications'
3. Enter a new URL

**Expected behavior**
For the field to be as responsive as the other fields

**Additional context**
https://github.com/CDH-Studio/UpSkill/pull/720

BenJeau:
#### Changes introduced

- Made skeletons the same height as empty form views
- Renamed `NameOptions` to `nameOptions` for the prop
- Made tab error checking happen onValuesChange instead of onFieldsChange (this makes the error checking seem a bit behind in the qualifications form - but I think its a good trade off for getting the forms/url field running smoothly)

#### Related issue(s)

patch / fixes #821

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible

### DIFF
--- a/services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
+++ b/services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
@@ -69,7 +69,7 @@ const EmploymentDataFormView = ({
   /* Component Styles */
   const styles = {
     skeleton: {
-      minHeight: "400px",
+      minHeight: "658px",
       maxWidth: "900px",
       background: "#fff",
       padding: "30px 30px",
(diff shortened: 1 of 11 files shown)


services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
```diff
@@ -69,7 +69,7 @@ const EmploymentDataFormView = ({
   /* Component Styles */
   const styles = {
     skeleton: {
-      minHeight: "400px",
+      minHeight: "658px",
```
dynamic11, line 72: please move these to separate PR. I like the change and i don't want to revert it in case the URL lag fix needs to be reverted

dynamic11 (approved):

BenJeau:
It's easier for me to create new branches/PR, I'll close this PR

PR #839, opened 2020-09-09
[Fix] Url field lag

### LINKED ISSUE #821: The URL attachment field input is not very responsive

BenJeau:
#### Changes introduced

- Made tab error checking happen onValuesChange instead of onFieldsChange (this makes the error checking seem a bit behind in the qualifications form - but I think its a good trade off for getting the forms/url field running smoothly)

#### Related issue(s)

patch / fixes #821

NA

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible

### DIFF
--- a/services/frontend-v3/src/components/profileForms/qualificationsForm/QualificationsFormView.jsx
+++ b/services/frontend-v3/src/components/profileForms/qualificationsForm/QualificationsFormView.jsx
@@ -171,8 +171,9 @@ const QualificationsFormView = ({
     );
   };
 
-  const onFieldsChange = () => {
+  const onValuesChange = () => {
     findErrorTabs();
+    checkIfFormValuesChanged();
   };
 
   /*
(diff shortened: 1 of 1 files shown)


BenJeau:
split of #837

dynamic11:
with this change we get a situation like this: field has error but tab doesn't show it. once we save they both show the error.

image

dynamic11 (approved):

PR #844, opened 2020-09-10
[Fix] Moved back to visible prop

### LINKED ISSUE #843: Roles in admin user table are all shown
**Describe the bug**
The keycloak roles in the admin dashboard are showing all of them for each users regardless of their actual roles

**To Reproduce**
1. Go to 'Admin'
2. Click on 'User' section in the left menu
3. See all of users

**Expected behavior**
To show the appropriate roles for the users - and not all of them

**Screenshots**

image

BenJeau:
#### Changes introduced

- Used the `visible` prop instead of `visibility` prop for the `<Tag />` and `<Modal />`

#### Related issue(s)

fixes #843

introduced by #813 (@dynamic11 was there a reason you switched it to `visibility` (visibility sounds less appropriate compared to visible, since its just a boolean)? - Just did a quick diff of that PR, and you also changed the `visible` props for the Modal components which accepts only `visible` and not `visibility` https://ant.design/components/modal/#API - And now the modals in the admin dashboard are also not working)

image

### DIFF
--- a/services/frontend-v3/src/components/admin/categoryTable/CategoryTableView.jsx
+++ b/services/frontend-v3/src/components/admin/categoryTable/CategoryTableView.jsx
@@ -112,8 +112,8 @@ const CategoryTableView = ({
     ),
     onFilter: (value, record) =>
       record[dataIndex].toString().toLowerCase().includes(value.toLowerCase()),
-    onFilterDropdownVisibleChange: (visibility) => {
-      if (visibility) {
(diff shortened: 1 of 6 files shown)


dynamic11:
I changed `visible` to `visibility` because in the forms `visible` can either receive either boolean or enum (connections, private, public). The variable name `visible` makes it sound like it can only be a boolean. However, I didn't mean to change the admin tables. That was a mistake.

dynamic11 (approved):

Issue #836, opened 2020-09-09
Node.js Security Checklist


dynamic11:
# Node.js Security Checklist
This is a checklist to secure Node.js application, Please implement each item and leave a description.

## Useful links:
- https://blog.risingstack.com/node-js-security-checklist/
- https://medium.com/swlh/3-security-pitfalls-every-react-developer-should-know-c04715b876b5
- https://www.simform.com/react-security-vulnerabilities-solutions/
- https://expressjs.com/en/advanced/best-practice-security.html

### Security HTTP Headers
we should use `Helmet` to set our headers https://www.npmjs.com/package/helmet for backend
https://www.twilio.com/blog/2017/11/securing-your-express-app.html

- [x] **Strict-Transport-Security enforces secure (HTTP over SSL/TLS) connections to the server**
Description: implemented for backend

- [x] **X-Frame-Options provides clickjacking protection**
Description: implemented for backend

- [x] **X-XSS-Protection enables the Cross-site scripting (XSS) filter built into most recent web browsers**
Description: implemented for backend

- [x] **X-Content-Type-Options prevents browsers from MIME-sniffing a response away from the declared content-type**
Description: implemented for backend

- [x] **Content-Security-Policy prevents a wide range of attacks, including Cross-site scripting and other cross-site injections**
Description: implemented for backend

- [ ] **HTTP/2**
Description:

### Authentication
- [x] **Authentication is handled by keycloak for our application.**

#### Cookie Flags
- [x] **Secure**: this attribute tells the browser to only send the cookie if the request is being sent over HTTPS.
Description:

- [x] **HttpOnly**: this attribute is used to help prevent attacks such as cross-site scripting, since it does not allow the cookie to be accessed via JavaScript.
Description:

#### Cookie Scope
- [x] **domain** :  this attribute is used to compare against the domain of the server in which the URL is being requested. If the domain matches or if it is a sub-domain, then the path attribute will be checked next.
Description:

- [x] **path**:  in addition to the domain, the URL path that the cookie is valid for can be specified. If the domain and path match, then the cookie will be sent in the request.
Description:

- [x] **expires** :  this attribute is used to set persistent cookies, since the cookie does not expire until the set date is exceeded.
Description:

#### CSRF
- [ ] **CSRF Token**
Description:

### Data Validation
- [ ] **XSS**:  filter/sanitize user input
Description:

- [x] **SQL Injection**:  filter/sanitize user input
Description: Prisma protects against SQL injection
[...]

PR #846, opened 2020-09-10
[Feat] Specified config for sessions

BenJeau:
#### Changes introduced

- sets the following to our cookie session management
  - **name** - sets the name of the cookie, we should just change the default one so people wouldn't know we're using express (Environment variable: `SESSION_NAME`)
  - **secure**
  - **httpOnly**
  - **expires** - one hour after creation
  - **domain** - should be specified to the domain of the application (Environment variable: `COOKIE_DOMAIN`)
  - **path** - can be anything, from my understanding (Environment variable: `COOKIE_PATH`)
  - **sameSite** - should be true (I think), since in our case both are from the same domain (true if production - on servers)

> __Note:__ the envs only need to be set on the server

#### Related issue(s)

cookies part #836

### DIFF
--- a/services/backend/src/auth/keycloak.js
+++ b/services/backend/src/auth/keycloak.js
@@ -22,11 +22,22 @@ const keycloak = new KeycloakConnect(
   }
 );
 
+const expiryDate = new Date(Date.now() + 60 * 60 * 1000);
+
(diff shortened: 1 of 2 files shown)


dynamic11 (approved):
👍

PR #892, opened 2020-09-21
[Feat] Migrate github actions to jenkins

### LINKED ISSUE #886: Port over Github Actions to Jenkins
Now that we are hooked into ISEDs Jenkins CI/CD system, we should also move over our linting and testing over

mo-5:
#### Related issue(s)

closes #886

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


mo-5:
Should be good for review. Please turn off linting on github action as well as i18n since I have moved them to jenkins. Also please make jenkins a requirement. I have left out the backend testing until its resolved, this way the jenkins build will not take a longer time.

dynamic11:
I tested removing the "required" tests and its works! @MohamedRadwan

.github/workflows/linter.yml
```diff
@@ -1,102 +1,102 @@
-name: Linter
```
BenJeau, line 1: Feel free to delete this file, we can always have access to it via the git history

Jenkinsfile
```diff
@@ -11,14 +11,60 @@ pipeline {
...
+                }        
+
+                stage('frontend') {
+                    steps {
+                        dir("${FRONTEND_DIR}") {
+                            sh """
+                                unset NPM_CONFIG_PREFIX
+                                source $NVM_DIR/nvm.sh
+                                nvm install "12.6.0"
+                                npm i yarn -g
```
BenJeau, line 42: would it be possible to do this only once (like once for the backend and the frontend)?

BenJeau (commented):
Seems good, removed the required old github actions - just seems to take a little while to lint it

BenJeau:
also another small question, you enabled parallel stages for linting, can/should we re-enable them for the building of the app?

Jenkinsfile
mo-5: From what I understood from derek is that you need to set it up in every sh or else it would reset. We could merge both backend and frontend together but then we wont know what failed the build. You would have to go through the text to find out.

mo-5:
> also another small question, you enabled parallel stages for linting, can/should we re-enable them for the building of the app?

I could test it out and see but I think it will still fail

mo-5:
Not sure why the Jenkins build is failing now it was working last week. interesting. I will look into it on Wednesday.

mo-5:
All done moved linter into one task so that we only set up node js once instead of twice. Please take a look and let me know.

BenJeau (approved):

mo-5:
Alight so I compared speeds and its actually faster as having backend and frontend separate and running in parallel so I will make it that way

PR #904, opened 2020-09-22
[Fix] Moved the "resetting" of form after saving to db

### LINKED ISSUE #901: When the form doesn't save, the user should be able to save the form again
**Describe the bug**
When triggering a 403 or 422 on the backend server when trying to save some form information (which happens when the server will not accept the request because it doesn't pass server side validation) or any other weird error, the form unsaved status is reset and the form thinks it has been saved and the user will not be able to "resave" the form data unless they make a modification. 

**To Reproduce**
I don't have an exact way to reproduce this, but it can often happen

**Expected behavior**
For the user to be able to "resave" the form after the error

BenJeau:
#### Changes introduced

- The resetting of the state of the form was moved after successfully saving to the database in the backend (modified onSave functions)

#### Related issue(s)

fixes #901

### DIFF
--- a/services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
+++ b/services/frontend-v3/src/components/profileForms/employmentDataForm/EmploymentDataFormView.jsx
@@ -308,9 +308,9 @@ const EmploymentDataFormView = ({
     form
       .validateFields()
       .then(async (values) => {
+        await saveDataToDB(values);
         setFieldsChanged(false);
         setSavedValues(values);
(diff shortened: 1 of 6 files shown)


dynamic11:
shouldn't it save then set the state? what is the benefit of this change?

BenJeau:
Yes, that's what my change is doing. We were updating the state before actually saving it to the database

dynamic11:
my bad. I read it the other way around lol 
I will run and test tomorrow morning

dynamic11 (approved):

PR #919, opened 2020-09-27
[New] Add webpack analyzer

dynamic11:
#### Changes introduced
- Add webpack analyzer to see the contents and size of webpack
- run 'yarn analyze' to run the analyzer
- visit localhost:3031 to see the visualization

this is a screenshot of the analyzer
image

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


BenJeau (approved):

BenJeau (changes requested):
It's failing in Jenkins for some reason - seems related to this change - you may need to add it to the deps and not devDeps, not sure image

dynamic11:
> It's failing in Jenkins for some reason - seems related to this change - you may need to add it to the deps and not devDeps, 

hmmmm.....you are right I forgot that the razzle plugin is also there for prod. Let me think of a solution. (i really don't think we should have it in production...let me know what you think) 

@BenJeau if you have any idea on how to have "development only" plugins let me know. I can't seem to find it in the documentation

BenJeau:
@dynamic11 not sure, but I have a feeling it doesn't matter if we just put it in the deps instead of devDeps, I don't think it has an influence on the react bundle size, just our docker image will be slightly bigger

BenJeau:
For reference, I tried reducing the size of antd.less with https://github.com/ant-design/babel-plugin-import, with no success, maybe others will be able to do it

BenJeau (approved):
Sorry, forgot about this PR - was waiting to see if it builds successfully on Jenkins - LGTM :+1:

dynamic11:
snyk is now failing

BenJeau:
mhh, should we merge it still?

PR #920, opened 2020-09-28
[Feat] Reduce bizchart import size

BenJeau:
#### Changes introduced

- Removed `<Geom />` components and used their respected geometry components, alongside with the other chart components
- Importing them like that also removes the old "hacky" way of importing them since bizchart does not support SSR (razzle)

#### Related issue(s)

https://github.com/CDH-Studio/UpSkill/pull/905

Before (size of bizchart in our bundle):
Screenshot from 2020-09-27 23-19-39

After (size of bizchart in our bundle):
Screenshot from 2020-09-27 23-18-05


dynamic11:
Wow really impressive savings! @BenJeau

dynamic11 (approved):

PR #928, opened 2020-10-01
[Fix] Lang property not defined in document head error for accessibility

### LINKED ISSUE #927: Lang property not defined for document head
**Describe the bug**
- Need to define lang property for the document head.

**To Reproduce**
1. Go to the front page.
2. Click on chrome dev tools.
3. Switch to Google Lighthouse.
4. Run the desktop report and see the accessibility issue.

**Expected behavior**
- No error if lang set to default "en".

**Screenshots**

**Additional context**

SukhsimranpreetSekhon:
#### Changes introduced

- Fixes lang property not being set issue from Google Lighthouse in the document head (accessibility error).

#### Related issue(s)

closes #927 

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [x] It is translated in both language (with no hard coded text) 
- [ ] Has been tested on IE
- [x] It is accessible


SukhsimranpreetSekhon:
I checked for every page if it takes the error away, it did.

BenJeau:
Should the lang attribute always be "en", even when the content is in french? If we set it statically, we should just modify the index.html, but if it need to change, react-helmet would be useful. You can get the current language/locale from redux

mo-5 (approved):

MohamedRadwan:
what happens when the language is changed to french?

mo-5 (changes requested):
what happens when the language is changed to french ? ops I just saw ben already made a comment on that

BenJeau (approved):
:100: LGTM

mo-5 (approved):

PR #935, opened 2020-10-03
[Feat] Added logo loading when loading React bundle

BenJeau:
#### Changes introduced

- If the user loads the application for the first time since a little while, it would show the logo - while the bundle is loading (not visible when navigating the app, but rather on its first load)
- Inspired from twitter

Peek 2020-10-03 14-46


mo-5 (approved):
I like this

PR #940, opened 2020-10-08
[New] GEDS sync modal

dynamic11:
#### Changes introduced
- added new GEDS sync modal
- individually sync fields from GEDS
- compare saved data with profile from GEDS

#### Related issue(s)
closes #939 

#### Screenshots (if applicable)
syncbtn

#### Checklist
If the database has been modified:
- [ ] Update database diagram 
- [ ] Create new migration 

If API endpoints has been modified:
- [ ] Update backend documentation 
- [ ] Create/Update validators for the API endpoints 
- [ ] The API endpoints are properly secured with keycloak 

If the UI has been modified:
- [ ] It is translated in both language (with no hard coded text) 
- [ ] Has been tested on IE
- [ ] The modifications are tab friendly
- [ ] It is accessible

Issue #939, opened 2020-10-08
Change the GEDS sync modal to allow user to choose element to update


dynamic11:
**Is your feature request related to a problem? Please describe.**
The current GEDS sync button syncs the entire profile without providing the user with control over what elements to update.
Example: user wants to update "Job Tittle" but not the "First Name"

**Describe the solution you'd like**
- When a user clicks on sync button in "primary info form" a modal opens up to compare saved data vs GEDS data. 
- The user is able to choose which element to sync

PR #942, opened 2020-10-08
[Fix] Remove Unsued GEDS Sync Endpoint

dynamic11:
#### Changes introduced
- removed geds sync endpoint as it is no longer necessary with the new GEDS sync modal
- should be merged after #940  

#### Related issue(s)
related to #939


BenJeau (approved):
Looks good, just merge your other GEDS PR first

PR #954, opened 2020-10-14
[Feat] ant design locale

BenJeau:
#### Changes introduced

- https://ant.design/docs/react/i18n
- Made ant design components translate to French when toggling language (and vice-versa for English)
  - But this refreshes the pages... as a side effect, not sure why

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


dynamic11 (approved):
tested and it works!

PR #955, opened 2020-10-14
[Feat] Improve theming throughout the application

BenJeau:
#### Changes introduced

This is a suggestive change, I made a lot of changes to the styling (please let me know what you think, I wouldn't mind if we don't merge the entirety of this PR, but we should at least add a few things from it)
- The navbar is now white to put less emphasis on it
- Everything is rounded to have a uniform, consistent, and unique look from the default stock ant design look
- Added subtle shadows a bit everywhere
- Change the header styling
- Made side menus consistent in admin and edit profile
- Reverted to the old logo, since there's less "roundiness" in this PR
- Footer has transparent color, so its more seamless
- Headers all have an icon

It also includes the following changes, that are not strictly towards the theming of the app
- Added GCdirectory sync to the side menu in edit profile
- Standardize the spacing of icon and text with `<span/>`
- Added subtitle prop for headers (to be used in profile and results page)
- updated login button on landing page

image
image
image
image
image
image
image
image
image
image
image

#### Checklist
If the UI has been modified:
- [ ] Has been tested on IE
- [x] It is accessible


mo-5:
I like these changes

BenJeau:
I'm open to any opinions, I know for a fact the home page looks a bit barebones, not sure how to fix it

dynamic11:
This issue was not casued by this PR but we should fix it in here. it happens in the education form on smaller screens (probably caused by an old change of mine)

image

dynamic11:
My vote is to remove the transparency of the nav on the search page
1) it makes the code simpler
2) I think it looks better

BenJeau:
I don't mind removing the transparency, it was just for the home page, so its not used a lot, and the home page looks a bit weird, so maybe it will fix it

BenJeau:
I will solve the conflicts, fix the linter, and cleanup this branch (I think I will create a seperate css file stricly for tabbing/accessibility , one for customization of ant design, and one for generic overall classes used in the code, so the css is clearer)

> **Edit:** done

BenJeau:
I will also fix the green background border radius difference in the qualifications form 
> **Edit:** this is now fixed

BenJeau:
@dynamic11, when you have time, could you create a .png of the logo? for logo consistency, unless you figure out a way to create an svg that doesn't rely on the font family (converting it to shapes I think)

BenJeau:
Had to rebase the branch with dev, because of all of the conflicts, explains why the commits are re-writen above

BenJeau:
small note, haven't tested in IE, it may not work. If it doesn't, I have an idea why (similar to the tabbing IE fix with classes seperated by commas)

BenJeau:
@dynamic11 the spacing doesn't affect functionnality if we remove the `marginBottom: -50`, so should we just remove it?

dynamic11:
Yea. I think I have the fix. Let me test and push tonight

BenJeau:
I think this branch is good to go, @dynamic11?

dynamic11:
I think this branch is good to go! I requested @MohamedRadwan and @SukhsimranpreetSekhon for code review since they are the "third party" with regards to this PR.

dynamic11:
@BenJeau I just tested IE and app wont load. its stuck on that logo splash screen

image

BenJeau:
Mhh @dynamic11 can you check if dev works in IE? (to make sure that that error comes from this branch or not)

dynamic11:
> Mhh @dynamic11 can you check if dev works in IE? (to make sure that that error comes from this branch or not)

Dev on openshift works. I'm about to test Dev locally to make sure its working there.

services/frontend-v3/src/components/layouts/landingLayout/LandingLayout.jsx
```diff
@@ -1,13 +1,31 @@
...
-  return <LandingLayoutView />;
+  /**
+   * Random Picture Select
+   *
+   * select a random picture for landing page
+   */
+  const randomPictureSelect = () => {
+    const imageOptions = [
+      backgroundOptionOne,
+      backgroundOptionTwo,
```
mo-5, line 21: just wondering why are we selecting a random image for the background?

mo-5 (approved):
LGTM

dynamic11:
found this issue with padding
image

dynamic11:
IE issue resolved

BenJeau:
the `custom_antd.scss` was overriding the changes in `index.scss`, so I moved the `index.scss` down and it fixed that checkbox issue, should be good now

services/frontend-v3/src/components/layouts/landingLayout/LandingLayout.jsx
dynamic11: honestly it was just a little fun thing we did cause we had a few left over designs. i can always remove it if its causing issues

mo-5:
Since we are looking to reduce the size I think we should remove all unused images. I also think instead of randomly picking an image we should pick one and delete the rest.

BenJeau:
Looking at the build folder in openshift, it only includes images used in the application.
image
Also, this is not included in react bundle, as it is making seperate API calls for it, which caches them on the client afterwards. So I think we're fine, unless we want the client to only make the same API call for the landing page image, but then again, that image is less than 20kb each and that API calls is asynchronous with the call for the react bundle

dynamic11:
I'm merging this bad boy in in the next 5 min. let me know if you want me to wait. 👿

Issue #975, opened 2020-10-21
Make frontend react bundle smaller


mo-5:
What's everyone's opinion on this? It seems like a pretty large package and something we would need to maintain. We are only using it in a few places. 

The advantages of removing this are: 
- Speed up Jenkins build and local build (1 less package to install) 
- Won't need to maintain when they release new updates 
- Speed up the application for the client ( the VPN is extremely slow so making the build package as small as possible should be a priority) 

The disadvantage is we need to convert all the functions to vanilla js.

Let me know what you guys think.

mo-5:
Also, I am not sure if we are using the following packages 
- serve ( I think this is for webpack / maybe moved to dev dependencies)
- query-string

Should these be moved to dev dependencies: 
- prop-types
- antd-dayjs-webpack-plugin
- webpack-bundle-analyzer
- webpackbar

mo-5:
The main concern with this is the performance on the VPN. Something we need to keep in mind which is why I think this is a prod blocker. If the performance is not good we will lose interest very fast.

BenJeau:
- `serve` is used to serve the built production application, so I don't think that one should go in the dev deps
- `query-string` is used to extract the vars from the url, ex: `?classifications[]=asfadsf&searchValue=sdfdsfd` in the results page

For the others, I think you can put them in the dev dependencies, but I think our current linter settings will complain, so we would need to modify it.

Quick question on the concern for the performance on the VPN, is this related to the speed of Jenkins is being built, or the size of the react bundle delivered to the client web browser? From this PR #930, the bundle is less than 1 MB, which imo is small enough and then gets unzipped on the client. Or is the network in inspect element saying otherwise? I don't understand why the performance on the VPN is slow, maybe its simply related to the fact that our application is outside the network?

BenJeau:
Wonder if #991 would make it smaller, since it doesnt rely on `craco` nor `react-scripts`.

> Edit: nevermind that doesn't makes sense, since those are just tools and are not incorporated inside the react build

BenJeau:
Kind of a regression here #1015, since I removed the ant design day.js webpack plugin (it was crashing the application, I would rather the app to be big than to crash right now)...

mo-5:
This issue can be closed since most of the big modules have been tackled

PR #988, opened 2020-10-23
[Feat] Added hot reload and moved things to devDeps

BenJeau:
#### Changes introduced

- Cleans up a bit the docker compose file
- Added hot refresh for local development (doesn't reload the page, it just refreshes the components)
- Clean up the craco config file to only include things according to the environment and moved things to the dev dependencies

#### Related issue(s)

NA, #975

Hot refresh example
Peek 2020-10-23 12-44

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,8 +6,7 @@ services:
       context: ./services/frontend-v3
       dockerfile: Dockerfile.local
     ports:
-      - "3000:3000" # port razzleJS
-      - "3001:3001" # port razzleJS
+      - "3000:3000" # port craco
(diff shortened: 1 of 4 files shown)


mo-5:
I like this. Could you also update the docker files to use italent instead of upskill for folder name ? Just so we switch out of the old name

mo-5 (approved):

BenJeau:
@MohamedRadwan I will do the name change throughout the repo in another PR, just in case it frigs up something (same with the `web-ant` container name in docker compose, the image names, and the folder name `frontend-v3` service under `services`)

PR #1000, opened 2020-10-25
[Feat] Switch to neutrino

### LINKED ISSUE #991: Switch from craco to neutrinos.js
I'm not sure how I didn't see this before, but react themselves recommend neutrino.js in their [README](https://github.com/facebook/create-react-app#whats-included). We should switch to it, since it does not rely on `react-scripts` and it will be less finicky updating our frontend packages. In addition, I'm not even sure if `craco` will support the latest version of `react-scripts`

Switch to neutrino.js

We've tried `react-app-rewired` with `customized-cra` in the past, then switched to `razzle.js`, and now we are using `craco`.

**Additional context**
https://github.com/gsoft-inc/craco/issues/205

BenJeau:
#### Changes introduced

- Removed everything related to craco
- More deps (but smaller node_modules) - more control over the versions of our deps
- Reduces the size of the build by a little
- Added css minimizer and js uglifier

> **Note:** wasn't able to apply the `babel-plugin-import` multiple times, https://github.com/neutrinojs/neutrino/issues/1646, hope to fix the related bug so we can have multiple instances of that plugin

#### Related issue(s)

closes #991

Screenshot from 2020-10-25 01-40-59

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


BenJeau:
At first glance it may seem like there are more dependencies added to the project, since there's more deps in the package.json, but if you look at the diff for yarn.lock, you'll see that the project is actually dependent on less dependencies since we don't depend on react-scripts nor craco (which has a lot of deps)

BenJeau:
alsooo, just realized that `@antv` in the photo above is used by the bizcharts package... so if we remove that, our react bundle should be smaller. Here's the bundle without bizcharts, almost 2MB less

Screenshot from 2020-10-25 01-54-07

@MohamedRadwan I know you'll be interested in this discovery :stuck_out_tongue: 

In comparison, the lib I suggested here https://github.com/CDH-Studio/I-Talent/issues/950 (https://www.npmjs.com/package/react-apexcharts) depends only on prop-types which has an unpacked size of 26KB compared to the 6.12MB of bizcharts. (6.12MB for only four charts in the admin dash is intense).

Just realized that I think I can apply the babel plugin import for ant design icons so it would just keep what is needed, juste like `antd/es`. Core-js is from the polyfills to support IE, so we can't remove that, but I think validator.js can be shrinked with the same technique, I'll try tomorrow or Monday

mo-5:
Oh wow, that's a great change. I guess with apex charts we will notice huge performance improvements. Also for the validator, I noticed that we can do es import to shrink the package.

BenJeau:
Just made validator.js smaller

141KB -> 7KB (parsed)
19KB -> 2KB (zipped)
Screenshot from 2020-10-25 14-23-52
Screenshot from 2020-10-25 14-23-56

mo-5:
Also I think we can reduce @ant-design using es import

BenJeau:
just tried using the same technique, but its not dropping the size, it seems like its already using es (for ant design components and icons)?

> **Edit:** seems like I can only apply it on one package actually? once I made it work with validator.js, the one for ant design wouldn't work anymore.... this is strange
> 
> Here, it would only apply it to validator.js, and not the ant design and it would inherently break ant design...
> image

BenJeau:
The only blocker for this PR is the runtime environment variables, not sure how to get them working with neutrino, but I think I'm close (and the multiple `babel-plugin-import` that is not working, technically not a blocker, can be fixed later on)

dynamic11:
image

dynamic11:
image
image
page fails to load randomly with following error (when filling out forms)

BenJeau:
Not sure why there's a warning about the hot reload, I can look into it, but I don't think it should cause any bug. For keycloack, I think thats related to an update, not 100% sure. But in your second image, it seems like the env.js file isn't getting created, but it worked for me yesterday.

I'll look into it

BenJeau:
I fixed the local dev env issue, and not getting that keycloak error. Still looking into the react hot loader warning, but it should still work fine...

dynamic11 (approved):
tested in chrome. it works

PR #1001, opened 2020-10-25
[Fix] Jenkins loadsh

### LINKED ISSUE #999: i18n:check fail due to lodash-es
**Describe the bug**
please see attached screenshot. i see this error when I go to run the check. this blocks jenkins tests aswell

**To Reproduce**

**Expected behavior**

**Screenshots**
image

**Additional context**

mo-5:
#### Related issue(s)

closes #999

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,7 +27,7 @@ pipeline {
                     steps {
                         dir("${FRONTEND_DIR_I18}") {
                             sh 'npm init -y'
-                            sh 'npm i lodash'
+                            sh 'npm i lodash-es'
                             sh 'node check'
                         }
                     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,7 +27,7 @@ pipeline {
                     steps {
                         dir("${FRONTEND_DIR_I18}") {
                             sh 'npm init -y'
-                            sh 'npm i lodash-es'
+                            sh 'npm i lodash'
                             sh 'node check'
                         }
                     }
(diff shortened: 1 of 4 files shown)


BenJeau:
Looks good, but it is still failing since in the check and cleanup files, you can't use `import ...`, you need to use `const ...` like in the backend to import packages

mo-5:
I have a question about this. Do you know if this gets added to the build package? like the module files? Since if not we could just go back to lodash regular import @BenJeau Do you know?

BenJeau:
Mhhh not 100% sure. But I know as a matter of fact, you know how you seperated each lodash function with a different import, that wouldn't make a difference here. Also, with the babel-plugin-import that is being used for ant design and validator.js, it can be used for lodash-es and we could just have it so we import normally and let babel take care of splitting it in different imports. I'm just having trouble getting multiple instance of that plugin to work, but according to babel and the issues/readme of the plugin, it should be feasible.

mo-5:
I see because we cant use lodash-es for the i18 checks. I will modify that to regular lodash.

BenJeau:
Ohhh, is it because for lodash-es, you need to import it with import because that is the es format?

mo-5:
Yeah

BenJeau (approved):
The only issue I see with this is that we won't be able to run it locally, since lodash is not specified in our package.json - I'll accept it since Jenkins is more important than running it locally. I think there's a linter issue now - you can fix it in this branch if you want

image

mo-5:
Yeah, I was thinking about the local thing. I am not sure what's the best way to approach this. I was thinking of just creating a separate package.json file for it.

BenJeau:
I think you could add lodash to the package.json and it would work? I don't it would add it to the react-bundle, as weird as it sounds - but you can for sure just add another package.json (maybe one at the root of the repo?)

PR #1002, opened 2020-10-25
[Feat] Simplify ant-design custom styling

BenJeau:
#### Changes introduced

Just figured out that there's theming variables that can be overridden.... (which are these ones https://github.com/ant-design/ant-design/blob/master/components/style/themes/default.less)

- Modify the `ant-design` theme variables instead of creating a big scss override file specifying the ant design css class names

#### Related issue(s)

NA

NA

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible

### DIFF
--- a/services/frontend/src/antdTheme.js
+++ b/services/frontend/src/antdTheme.js
@@ -19,5 +19,5 @@ module.exports = {
   "@card-actions-background": "transparent",
   "@card-head-height": "55px",
   "@list-item-meta-margin-bottom": "0",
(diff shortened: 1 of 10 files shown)


dynamic11 (approved):
tested in chrome and it looks good! I also tested in IE but app wont get past load screen (i dont think it is related to this change). I think the babel pollyform is misplaced again (will look into it)

PR #1004, opened 2020-10-26
[Update] nodejs in docker

mo-5:
#### Changes introduced

update to 12.19.0 please test locally to make sure it works. 

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible

### DIFF
--- a/services/backend/Dockerfile
+++ b/services/backend/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.18.3-alpine3.11
+FROM node:12.19.0-alpine3.11
 
 ENV TZ=America/Toronto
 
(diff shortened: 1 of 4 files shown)


mo-5:
Also, I just noticed if we moved to npm for our node js image will shrink by ~half 

https://github.com/mhart/alpine-node

BenJeau:
But it seems like we would need to revert to an older version of node.js? I think it would be v6?

mo-5:
I don't think these are referring to the versions of the node. 0.10.48 would be a pretty old version. I think it's an internal versioning.

BenJeau:
But I think it is, looking up online they are valid node.js versions, the latest v0.12 release was in 2016 and v4 was in 2018 (and looking inside of the Dockerfiles they are using, it is specifying that node version). Its weird that they have it there

BenJeau (approved):
Works for me locally

PR #1006, opened 2020-10-26
[Fix] migrate to apex charts

### LINKED ISSUE #950: Move from Bizcharts to Chartjs-2
Bizcharts documentation is all in Mandarin and may the charts part of our code may not be well understood by maintainers later on... I just saw that the covid19 Ontario website is using ant design for the table and ApexCharts for its charts. At least their docs are in english and the design is not that bad

We should probably move to ApexCharts https://github.com/apexcharts/react-apexcharts

**Additional context**
You can see which library they use by looking at the source code of the website

https://covid-19.ontario.ca/data

mo-5:
#### Changes introduced

The style is a little different. I can't find much customization for apex charts. However, the application is much lighter. 
#### Related issue(s)

closes #950

image

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


mo-5:
also, @BenJeau when I try to create the analyze graph this is the error I get 
image

mo-5:
I think it does make a big difference because bizchart seems to have a lot of peer dependencies

BenJeau:
For the analyze, you need to attach your shell to the frontend contianert

mo-5:
oh, I didn't realize that loool. Sounds good, thanks. 
image
I am not sure if it made any difference tbh. It seems like apex chart import is pretty large

BenJeau:
What lol, that's weird... I'll pull your branch and play with it a bit. But yeah, it does seem really big

mo-5:
Will Convert to chart.js since it's smaller.

PR #1019, opened 2020-10-28
[New] Update from Bizcharts to ChartJs

mo-5:

image
image

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


mo-5:
before:
image

after:
image

mo-5:
I noticed this 
image

Does that mean ant- design can be imported using es ?

BenJeau:
@MohamedRadwan maybe, we could use the babel plugin thing, but its not working with more than just one lib right now, I tried it and don't remember if it went down that much. I think if we see `es` in the analyze means that we are using the `es`?

mo-5:
> @MohamedRadwan maybe, we could use the babel plugin thing, but its not working with more than just one lib right now, I tried it and don't remember if it went down that much. I think if we see `es` in the analyze means that we are using the `es`?

Sounds good, I think if we see the whole package including es that means we are not using it however if we only see es that means we are using es and not importing the whole package. Thats what I understand from this

services/frontend/src/components/admin/dashboardGraphs/DashboardGraphsView.jsx
```diff
@@ -138,7 +123,9 @@ const DashboardGraphsView = ({
               }
               loading={monthlyGrowth.length === 0}
             >
-              <Chart
+              <Chart type="line" data={monthlyGrowthData} options={options} />
+
+              {/* <Chart
```
BenJeau: Looks good, I would remove this comment and design wise for the charts, we can work on it after this is merged, unless you want to do it now

BenJeau (approved):

services/frontend/src/components/admin/dashboardGraphs/DashboardGraphsView.jsx
mo-5: I think we can work on the design after this merge just to test out chart js first

PR #1026, opened 2020-10-29
[Fix] Pass history to handleError

### LINKED ISSUE #800: Errors in the frontend are not properly caught - since Razzle.js
**Describe the bug**
Since the introduction of Razzle.js, we are now using the `<BrowserRouter />` instead of `<Router />` because the server side rendering (SSR) from Razzle.js. The history instance cannot be passed to the `<BrowserRouter />`, that component creates the history instance inside of it. We need to somehow figure out how to navigate the user to the error page from that `handleError()` function.

I see two possible solution to this problem:
1. Pass the history instance to the `handleError()` function everywhere that we call it
2. Somehow figure out a way get an instance of history with `<BrowserRouter />` (using the barebones `<Router />` will cause the page to not render because of SSR)
3. Just add a page reload after adding the `/error` path - the easiest solution

**To Reproduce**
1. Trigger an error
2. See that the url has changed to `/error` but the content has not been refreshed

**Expected behavior**
For the page to show the content of the `/error` page

BenJeau:
#### Changes introduced

Brings back what we used to have
- Pass history object to the `handleError` function
- Made the UnexpectedError page component use the same errorResult component as the other error components

#### Related issue(s)

I think this is good enough to close #800. Its not doing error boundaries, but I will open a new issue

Peek 2020-10-28 22-18

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


mo-5 (approved):

mo-5:
Please take a look at the linting errors

PR #1035, opened 2020-11-02
[Feat] Faster backend testing

### LINKED ISSUE #469: TopFiveX tests are flaky
**Describe the bug**
The TopFiveSkills, TopFiveCompetencies, and TopFiveDevelopmentalGoals tests are flaky

**To Reproduce**
1. Run the TopFiveX tests until they fail

**Expected behavior**
For the test to always pass

**Additional context**
https://github.com/CDH-Studio/UpSkill/runs/845858395?check_suite_focus=true
https://github.com/CDH-Studio/UpSkill/runs/845882203?check_suite_focus=true
and the other failed github actions

BenJeau:
#### Changes introduced

- Removes the need to have a dedicated testing `Postgres` or `Redis` database for our backend test which should make it super fast compared to before. This requires to mock Prisma, but I have limited knowledge in this field, so I am starting with only a few files and will move on once they work. I want the GitHub actions to pass instead of constantly fail like they do now, which is the main purpose of this PR (if its fine) or else it can be put on hold until the other tests have been fully ported over.
- Renamed old test to end with `.old.js` so they are ignored by jest (eventually should all be converted using the new way of testing the backend)

#### Related issue(s)

closes #469 as it removed them

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


BenJeau:
I decided to put this branch ready for review so that we can incrementally work on the tests in the backend instead of having one massive PR adding all of them

BenJeau:
@MohamedRadwan we could integrate Jenkins to run the backend tests after this PR is in dev, since it doesn't rely on any databases and the tests that have been converted are passing (you can look at the GitHub action and we could remove that, so now would be solely on Jenkins)

mo-5:
This is a great update. Do you want to just add the jenkins portion for the testing ?

BenJeau:
I can add it if you want, where should it be in the flow? parallel with the linting, with the linting, before the linting or after it?

mo-5:
After the linting, since the linting is faster and if that fails first then there is no point to continue

mo-5:
Sorry, can you actually add it as part of the parallel linting and i18 check? This way the building process will be faster.

BenJeau:
@MohamedRadwan Is the change what you had in mind (for jenkins)?

mo-5:
Yea that way we reduce the build time

BenJeau:
@MohamedRadwan please look at the last commit, I think I figured out how to setup node and then use that node instance throught the stages (it just needs to source that nvm bash file and it will use that specified node version). Feel free to approve this PR if you think its fine

mo-5:
LGTM!

mo-5 (approved):

PR #1039, opened 2020-11-04
[Fix] Search Filter - Results page accessibility

### LINKED ISSUE #1130: Accessibility Fixes & Additions for Results Page Search Filter
**Describe the bug**
- Fix Accessibility issues/errors from Google LightHouse Test.
- Add additional accessibility features and enhancements if necessary.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to 'Results Page'.
2. Click on the 'Dev Tools' option in settings.
3. Click on the 'Lighthouse' tab.
4. Run only the 'Accessibility' check.

**Expected behaviour**
- Even if the errors do not go away on the Lighthouse test or other accessibility testers, make sure issues are at least solved for user interaction. The testers may not always see the fixes/additions properly due to Ant Design's code configuration.

**Screenshots**
- None

**Additional context**
- None

SukhsimranpreetSekhon:
#### Changes introduced

- Added/Fixed aria labels and roles for the search filter form items.
- Fixed screen reader reading/output for filter items.
- Added additional aria attributes/properties to form items for tab/keyboard and screen reader-friendly interactions.

#### Related issue(s)

fixes #1130 

- Needs to be tested with an NVDA screen reader and user interaction to hear screen reading output and feel accessibility controls/highlights.

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If API endpoints have been modified:

If the UI has been modified:
- [x] It is translated in both language (with no hardcoded text) 
- [ ] Has been tested on IE
- [x] It is accessible


SukhsimranpreetSekhon:
Adding commits ongoing.

BenJeau:
Is this PR ready to go (I only see changes to the intl translation files)? if not I'll mark is as a draft

services/frontend/src/components/searchFilter/SearchFilterView.jsx
```diff
@@ -117,10 +158,33 @@ const SearchFilterView = ({
...
           >
             {locationOptions.map((value) => {
+              location =
+                value.streetNumber +
+                " " +
+                value.streetName +
+                ", " +
+                value.city +
+                ", " +
+                value.province;
```
BenJeau: 1. Since you are defining the content of the location, which is the same that is in the Option component, inside of Option, I would just have `{location}` instead of defining the location string twice

2. You need to declare the location variable, and you can concatenate a string like this
```suggestion
              const location = `${value.streetNumber} ${value.streetName}, ${value.city}, ${value.province}`;
```

or if you decompose the object in the param of the map loop, like this `({streetNumber, streetName, city, province, id}) => {...` it could be simplified to the following and you could use straight the variable `id` in the key of the `<Option />` component.

```suggestion
              const location = `${streetNumber} ${streetName}, ${city}, ${province}`;
```

services/frontend/src/i18n/en_CA.json
```diff
@@ -343,6 +343,14 @@
   "search.advanced.general.title": "General info",
   "search.advanced.skill.title": "Skills and talent",
   "search.empty.query": "Refine your search criterias to start finding talent!",
+  "search.filter.name": "Input Name",
```
BenJeau: I would rename this to `Person name` or something, to be more descriptive, it would be `Nom de personne` in French

services/frontend/src/i18n/en_CA.json
```diff
@@ -343,6 +343,14 @@
...
+  "search.filter.classification": "Select Classification",
+  "search.filter.location": "Select Location",
+  "search.filter.branch": "Select Branch",
+  "search.filter.skills": "Select Skills",
+  "search.filter.mentorship.skills": "Select Mentorship Skills",
+  "search.filter.any.mentors": "Search for Mentors?",
+  "search.filter.exfeeder": "Check for EX-feeder",
```
BenJeau: To keep consistency within the application, usually nothing is title case because of the Government standards, could you change it to sentence case?

BenJeau (commented):
Looks good, some small comments, and also if you are having any linting issues, running `yarn lint --fix` locally in the docker container of the respecting frontend/backend container should fix most of them (I see that Jenkins is not happy - the change for the concatenation will fix it)

services/frontend/src/i18n/en_CA.json
SukhsimranpreetSekhon: Fixed.

services/frontend/src/i18n/en_CA.json
SukhsimranpreetSekhon: Fixed.

services/frontend/src/components/searchFilter/SearchFilterView.jsx
SukhsimranpreetSekhon: Changed it to the second one, it made more sense and easier to follow for users who might update the code later on.

BenJeau (approved):

mo-5:
Is this good to be merged?

Issue #1017, opened 2020-10-28
IE is not working (because of arrow functions)


BenJeau:
**Describe the bug**
The application is not working on IE

**To Reproduce**
1. Go on IE and see the application favicon forever

**Expected behavior**
The application to work on IE

**Screenshots**
![image](https://user-images.githubusercontent.com/22248828/97451384-13f32000-190a-11eb-9387-8aa12cc3fe97.png)

**Additional context**
I think this is related to the terser uglifier (just need to modify the arrow config https://github.com/terser/terser#compress-options)

BenJeau:
not sure if I'll be able to touch this, this week @dynamic11 feel free to takeover - I know think its somehow related to neutrino, https://neutrinojs.org/packages/react/#polyfills, or maybe this https://neutrinojs.org/packages/web/#targets

MenaMachado:
It doesn't seem to be working in Edge either

BenJeau:
Oh really, that's bizarre, I will take a deeper look

BenJeau:
@MenaMachado I'm done my hours for the week (so I won't be able to touch this issue). Is prod something we want to accomplish this week? If so, I can work a bit more and work less for the following week

MenaMachado:
Next week is fine

On Nov 4, 2020, at 6:23 PM, Benoît Jeaurond <notifications@github.com> wrote:

﻿

@MenaMachado<https://github.com/MenaMachado> I'm done my hours for the week (so I won't be able to touch this issue). Is prod something we want to accomplish this week? If so, I can work a bit more and work less for the following week

PR #1045, opened 2020-11-04
[Fix] Update deps - Minimize react-bundle

BenJeau:
#### Changes introduced

- Added babelrc file, which allows for multiple instance of the `babel-plugin-import` (to shrink the react-bundle and not having to do the tree shaking manually in the code)
  - Removed manual tree shaking for lodash and validator.js
- Added `webpack` and `babel` plugins to reduce lodash size
- Added `webpack` ignore plugin to ignore the moment.js locales (doesn't seem to affect the application)
- Explicitly set browserlist in `.neutrinorc.js`, remove arrow functions from JS uglifier, explicitly set IE compatibility in babel plugins (but doesn't seem to still work on IE...)
- Updated frontend deps

I updated the deps in hopes to fix IE

> **Note:** I didn't update React to v17 since ant design is not supporting it right now, according to one of the issues, they will support it in the next major release, aka v5 (but neutrino supports it)

Before:
Screenshot from 2020-11-04 09-59-52
Screenshot from 2020-11-04 09-59-52

This PR:
Screenshot from 2020-11-04 09-57-53
Screenshot from 2020-11-04 09-57-53

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


BenJeau:
**Note:** the lodash fix is actually breaking the application... I need to look into it more

mo-5:
Are the issues in IE related to antd ?

BenJeau:
I don't think so, I think its related to neutrino.js, forgot to link the issue #1017 - I will jump back on this PR today, and plan on making it ready for review

BenJeau:
@MohamedRadwan Please test this branch locally (remove docker images/container and build from scratch) to make sure that it is not breaking anything major. It was breaking for me in the past, but just tried it today and it seems fine, not sure what changed

BenJeau:
Potential fix for #1055 (may be related to the update to axios)

mo-5:
I will test it out and see do you think we should add obfuscation for the javascript as part of this PR. I am not sure if that is just a one line code in the webpack configuration or not but it might be helpful to keep the frontend code secure.

BenJeau:
There seems to be a webpack plugin, https://github.com/javascript-obfuscator/webpack-obfuscator, but I'm not sure if we need obfuscation since technically our code is open source and it is only used internally

mo-5:
Yeah that is true

dynamic11:
I just tested this PR and seems to be working but IE is still not loading. I believe the PR title should be changed since IE still not fixed.

dynamic11 (approved):

PR #1090, opened 2020-11-17
[Fix] Cleaned up profile layout

BenJeau:
#### Changes introduced

- Fixes inconsistent padding/margins in the profile layout view on mobile and on desktop
- This makes the ex-feeder full screen width, or else there is side effects, but I think its not a bad tradeoff for this cleanup

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


dynamic11 (approved):
tested. LGTM

PR #1103, opened 2020-11-25
[Fix] Made search faster

BenJeau:
#### Changes introduced

- Does not reload the page when searching, just listens to changes to the URL

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


mo-5 (approved):

dynamic11:
I just noticed an issue with this fix. if we don't refresh the page, the advanced search option will not clear when trying a new search string in fuzzy search.

Please see GIF for example
saveissue

PR #1107, opened 2020-11-26
[Fix] Clearing search fields according to the fields being edited

BenJeau:
#### Changes introduced

- Made the search filter concord with the querySearch string from the URL (just clear the form before setting it to something)

#### Related issue(s)

related to #1107

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


dynamic11 (approved):
tested and it works 🥳

PR #1116, opened 2020-11-27
[Fix] Separate search result card into independent component

dynamic11:
#### Changes introduced
⭐ New Feature:
- separated the result card into a separate component than the result page
- cleaned up styling
- this feature should not change any functionality or styling

#### Related issue(s)
The change is not related to an issue. It is just an improvement (clean up) in preparation for a future feature to show matches for each card in a modal


BenJeau (approved):
Looks good, but I think it will conflict with #1039

Issue #1123, opened 2020-12-08
Improve fuzzy search 


dynamic11:
**Is your feature request related to a problem? Please describe.**
the current fuzzy search doesn't perform as well as expected. it has a lot of false positives and false negatives. with the merge of PR #1121 we can see the fuzzy search matches so it will be easier to optimize the search.

**Describe the solution you'd like**
- [x] searching 235 queen should bring up 235 queen addresses
- [x] searching EXfeeder should bring profiles with exfeeder set to true
- [x] branch acronyms should be searchable along side full branch name
- [x] add advance search options to fuzzy search such as "AND", "OR"
- [x] reduce threshold
- [x] search full name

dynamic11:
all items are finished

PR #1136, opened 2020-12-09
[Feat] generate an acronym for branch to be searchable

dynamic11:
⭐  **Changes introduced**
- when loading all profiles to feed into fuzzy search we now generate an acronym for the branch so it can be search by the acronym
- EX: profiles with `Human Resources Branch` can be found by searching `Human Resources Branch` or `HRB`

ℹ️  **Related issue(s)**
- addresses issue #1123


mo-5 (approved):

services/backend/src/core/search/util/getAllProfiles.js
```diff
@@ -293,7 +293,14 @@ async function getAllUsers(language, userId, request) {
 
     if (info.employmentInfo) {
       const employment = info.employmentInfo.translations[0];
-      info.branch = employment ? employment.branch : undefined;
+      info.branch = {};
+      info.branch.name = employment ? employment.branch : undefined;
+      info.branch.acronym = employment
+        ? employment.branch
+            .split(" ")
+            .map((word) => word.charAt(0))
```
BenJeau: I think maybe just the following is simpler (since strings are just an array of chars)?

```suggestion
            .map((word) => word[0])
```

BenJeau (commented):
Looks good, but could you also update the docs to reflect this change, for it to be an object instead of a string?

https://github.com/CDH-Studio/I-Talent/blob/3c05233e3ae15794c7e24ae17b9b30d3673e0be3/services/backend/src/docs/definitions.yml#L203-L204

services/backend/src/core/search/util/getAllProfiles.js
dynamic11: lol i always thought charAt was better practice but didnt know why. I just learnt its for support for IE7. lol since we dont need to support that and its the backend anyways then ill fix it 😂

dynamic11:
> Looks good, but could you also update the docs to reflect this change, for it to be an object instead of a string?
> 
> https://github.com/CDH-Studio/I-Talent/blob/3c05233e3ae15794c7e24ae17b9b30d3673e0be3/services/backend/src/docs/definitions.yml#L203-L204

will do! i actually put in a few other PRs that effect the docs so i should update those aswell

BenJeau:
Yes please do, I may have forgotten about that, you can do it in this PR if you want

dynamic11:
> > Looks good, but could you also update the docs to reflect this change, for it to be an object instead of a string?
> > https://github.com/CDH-Studio/I-Talent/blob/3c05233e3ae15794c7e24ae17b9b30d3673e0be3/services/backend/src/docs/definitions.yml#L203-L204
> 
> will do! i actually put in a few other PRs that effect the docs so i should update those aswell

I just realized that the data being returned by the API endpoint hasn't changed. in `cleanResults` i just grab the branch name. So i guess it wont need to be updated after all.

PR #1140, opened 2020-12-11
[Fix] Remove select all option for Skills and Mentorship Skills dropdowns

### LINKED ISSUE #1127: Remove select all option in tree select dropdowns
⚠️ **based on user testing**
- users don't understand that each category name in the tree select has a dropdown so they click the name of the category and it auto selects all items under it

ℹ️ **I propose**
- we remove the select all option

Mousto097:
⭐  **Changes introduced**
- Removed select all option for Skills and Mentorship Skills dropdowns in SearchFilter and SearchBar
- Set `selectable: false,`, `checkable: false,` and `disableCheckbox: true, ` for all parent nodes in dataTree
- updated antd to 4.9.4 as it contains `TreeSelect` updates
- prevent forms from submitting on 'enter' keypress

ℹ️  **Related issue(s)**
- Closes #1127


SukhsimranpreetSekhon:
Also need to fix: after selecting option value (example: "Specialization: Politics") and pressing the "enter" key, it will refresh the page instead of selecting the content.

SukhsimranpreetSekhon:
Also need to fix: category should not be a selectable/checkable item.

SukhsimranpreetSekhon:
> Also need to fix: category should not be a selectable/checkable item.

Maybe we can use the "disabled" prop to disable the category name from being selected.

SukhsimranpreetSekhon:
> Also need to fix: after selecting option value (example: "Specialization: Politics") and pressing the "enter" key, it will refresh the page instead of selecting the content.

Have to fix it up using the "onSelect" prop or there is something wrong with the form and form.item components sending the selected data through right away once the "enter" key is pressed.

dynamic11:
🐞 BUG FOUND:
on the advance search if you try to use the keyboard to select a skill and hit enter to make selection the form submits.

dynamic11:
Found the fix I think!
will push fix shortly! 🥳

dynamic11:
more fixes are coming! i need to update the rest

mo-5:
Is this pr ready for review?

dynamic11:
Yes it is 😁

mo-5 (approved):

mo-5:
Looks good

PR #1158, opened 2020-12-18
Split profile API

### LINKED ISSUE #1047: Split profile API endpoint
## Is your feature request related to a problem? Please describe.
The profile API endpoint is responsable for a lot of things, that endpoint is not a correct implementation of REST and feels like a custom graphQL mutation endpoint (in the way you can throw anything at and it will update, delete, or/and create entries). We should split it based on its content for the code to be cleaner, the responses to be shorter and thus faster. For example, the update and get profile endpoint has around 700-800 lines of codes each. Having it all in one endpoint also means that we have a lot of different logic for different parts of the profiles in one place, which means the code is highly coupled and that's not good. Seperating it would decouple our backend code and make more sense to new comers / maintainers.

> **Note:** this problem became more appearant as we build more and more features and cover all of the edge cases. If we don't do anything now the lines of code and complexity of the profile APIs will continue to grow. We are inherently not making it explicit what the profile endpoint can and cannot do

### Advantages
- More modularized code
- More testable code
- Faster API responses
- More conventional to the REST standard

### Disadvantages
- Need to change everywhere where we are calling the profile endpoint
- Need to make multiple API calls (whereas we are only making one now)
- Redundant code for the backend (but it makes it more testable, predictable, and divided) - just like what has been done to the options API in the summer

### What we have now (**_3_** APIs):
- **PUT** `/api/profile/:id`
- **GET** `/api/profile/:id`
- **GET** `/api/profile/private/:id`

#### Generic profile information
- **POST** `/api/profile` (getting rid of the `user` API)
- **DELETE** `/api/profile/:id` (getting rid of the `user` API)
- **PUT** `/api/profile/:id` (patches the values strictly belonging to the profile table)
- **GET** `/api/profile/:id` (gets the values strictly belonging to the profile table (which includes foreign ids to other tables))

#### Foreign ids of profile
- actingLevel
- careerMobility
- classification (groupLevel)
- lookingJob
- officeLocation
- securityClearance
- talentMatrixResult
- tenure

We would have similar APIs
- **POST** `/api/profile/{name}/:id`
- **DELETE** `/api/profile/{name}`

#### Visibility cards
- **PUT** `/api/profile/visibility`
- **GET** `/api/profile/visibility`

> **Note:** for the following I'm not sure about the **DELETE**

#### For each of the following
- competencies
- developmentalGoals
- educations
- experiences
- mentorshipSkills
- qualifiedPools
- relocationLocations
- secondLangProfs
- skills

We would have similar APIs
- **GET** `/api/profile/{name}`
- **PUT** `/api/profile/{name}` (with the ids in the body)

BenJeau:
#### Changes introduced

- Splits the profile API and removes the user API
  - Created the endpoints specified in the related issue 

This PR also uses Express middlewares to clean up the core functions, there are two
1. does not continue request if the requested profile is not the same user making the request
2. does not continue request if profile is not ACTIVE and not an admin

In summary, the modifications are the following
- Removed 3 user API endpoints
- Modified 3 profile API endpoints
- Created 37 profile API endpoints

> Note: this is still a work in progress, but I plan on having it done by the end of today. The only things left for this PR is the following
- for developmentalGoalsAttachment in developmentalGoals endpoint
  - [ ] update core logic
  - [ ] add validation 
- [X] Create API validators for the endpoints (should be easy since they just need to be moved)
- [ ] Finish integration with the frontend
- [ ] Create API docs for the endpoints

#### Related issue(s)

closes #1047 

#### Checklist
- [x] Create/Update validators for the API endpoints 
- [x] The API endpoints are properly secured with keycloak


BenJeau:
I worked a lot today and the past few days on this PR. @dynamic11 and @MohamedRadwan feel free to do what you guys want with this (close or continue working on it). It still needs some works, unfortunately, I thought I would be able to get it all done, but this is a big starting point.

What would need to be done is:
- Modify the frontend for all the calls to the profiles API (getting and updating values)
  - For getting the data, it should be easy, but the data that it returns is in this format for certain of the API `{ data: ..., updatedAt: ... }` instead of directly the data, so it will need some tweaking in the frontend (you can look in the last commit to see how to aggregate the data of a profile)
- Update API docs

mo-5:
Since I am working full time hours next week to catch up on the my missed hours I could take this pr over unless @dynamic11 wants to take it.

BenJeau:
Forgot to mention that the two first forms are fixed, but there are some bugs with accessing other people's profiles (but maybe fixing (in the frontend) viewing your own profile first will fix viewing others afterwards)

dynamic11:
@MohamedRadwan fell free to take this on

PR #1168, opened 2020-12-22
[New] Add print button to profiles

### LINKED ISSUE #215: Add "Print results" button to results page
Some managers may want to a physical copy of all resumes of the employees that match their search results, thus we need a way to easily print them out.

A button on the results page/in the results page sidebar that prints nicely formated resumes (sumaries of the profile page) of each applicant that matched the search criteria/is displayed on the results page.

mo-5:
#### Changes introduced

I added a print button to allow users to print their profile and other users' profile. Please note I removed the private connections list from the print that way the list remains private. 

There is one issue with this is that sometimes the cards cut and continue on the next page. Unfortunately, there is no good way to fix this since the cards are variable to the amount of information the user provides. 

#### Related issue(s)

closes #215

button location:
image
once you click on the button:
image
example:
image

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


dynamic11:
As seen in the image below the skills are all collapsed. I don't think this is very helpful. we can either expand it on print or generate a CSV/excel. I'm personally team CSV/Excel because it condenses the info and saves paper 😛 .

image

mo-5:
I think we could expand them on the print I would have to look more into it. There is a way we could create a pdf or csv/excel file instead of just a print screen but we would need to install another package.

mo-5:
I updated skills and mentorship to this. Since it is the only way to make them expand during command. Chrome stopped supporting event handlers during printing. 

profile:
image

print:
image



I have done this we will limit the number of skills a user can add to their profile. Please let me know if this will be okay.

mo-5:
image

Rizvida1 (approved):

PR #1202, opened 2021-01-05
[New] Update the functionality of the backend 

### LINKED ISSUE #1052: Backend Security Hotspot (SonarQube)
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**
image

**Additional context**

mo-5:
#### Changes introduced

These changes are based on the changes made by Ben in pr #1158. After a discussion with Ali, we have decided to make the changes done to the backend which would be a good update for the backend but keep the same endpoint as before. This will reduce the size of the change. This is done to avoid having bugs in the frontend.

#### Related issue(s)

I think I have also fixed this error (closes  #1052)

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


mo-5:
I have taken a lot of the things and wrote a little more on top and fixed a few issues. I was going to try to update the profile with the broken down code and call each code from profile.js however, it introduced a lot of errors. I have decided to keep the profile the same way to avoid bugs coming up later on. 


The tests pass locally I am not sure why they are failing on jenkins.

PR #1203, opened 2021-01-05
[Fix] GCdirectory bug

### LINKED ISSUE #1198: Geds not working
**Describe the bug**
we get 404 error. its probably due to a new url since migration.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

dynamic11:
🐞 Bug fix
- modified logic to prevent an array of different organizations
- modified Mary Doe's seed data to only be part of one org tree 
- improved formatting for org tree drop down with new message for empty tree
- updated tests

ℹ️ Related Issues
- closes #1198 

📸 Screenshots
image
image


mo-5 (approved):

dynamic11:
I found additional bugs. fixing as we speak

services/backend/src/database/schema.prisma
```diff
@@ -640,7 +640,7 @@ model User {
   qualifiedPools                QualifiedPool[]
   competencies                  Competency[]
   secondLangProfs               SecondLangProf[]
-  organizations                 Organization[]
+  organizations                 Organization?
```
mo-5, line 643: You will need to run a migration for this to take effect

dynamic11: will that wipe our data?

dynamic11: will it wipe our data?

mo-5: It shouldn't but there is a small chance it would cause an error if there is already data stored in the organization. You could test it locally and see. Run the migration after data is stored in the organization and see if it will delete them.

dynamic11: Will do!

dynamic11: so I just realized I never generated the migration files. I just added them. I also tested what you asked me to do and it fails. I think it has to do with the fact that marry doe has an array of orgs. I will do more tests tomorrow and try to find a fix.

mo-5: Oh yeah sorry that's what I meant but run a migration. Do you actually need to change the organization or could you set the default to an empty array? as far as I recall the "?" just makes it an optional field.

dynamic11: we can talk about it in the meeting today

dynamic11: ok so I confirmed this fix. I will need it to be as optional and not an array because if we leave it as an array then it will always return an array and that's not what we want. we also want to block anyone storing multiple org structs for each user.

When pushing this to the server the migration will fail because "marry doe" has multiple org structs. so we might have to go in to the db and delete the second one. @MohamedRadwan can you do that through pg-admin?

mo-5: Our seed values are not on either dev or prod. So marry doe is not on there.

dynamic11: perfect then this should not cause any issues! Im going to update the branch and merge. if anything happens we have the weekend to fix :)

PR #1205, opened 2021-01-05
[Fix] crash due to unknow location in search results card

dynamic11:
🐞 Bugs fixed:
- corrected logic to check if office location exists before attempting to render


services/frontend/src/components/resultsCard/resultProfileCard/ResultProfileCardView.jsx
```diff
@@ -202,7 +202,7 @@ const ResultProfileCardView = ({
     </div>,
     <div>
       <EnvironmentOutlined className="result-card-footer-icon" />
-      {user.branch ? (
+      {profile.officeLocatio ? (
```
mo-5: should that be officeLocation ?

dynamic11: omg i can believe what I just committed lol

mo-5 (approved):

PR #1211, opened 2021-01-08
[Fix] Fixes the current package warnings and errors

### LINKED ISSUE #1113: Frontend yarn related errors
**Describe the bug**
there are currently 372 errors and 3 warnings from our packages. 
**To Reproduce**
run yarn check

**Expected behavior**

**Screenshots**
image

**Additional context**

mo-5:
#### Changes introduced

- I moved the application back to npm since it doesn't build if there are errors with the dependency tree with is a good saftey net when we are updating packages.
- I updated keycloak-mock so we won't have any vulnerabilities https://github.com/MohamedRadwan/keycloak-mock/releases/tag/1.0
- This should make jenkins run slightly faster

#### Related issue(s)

closes #1113

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


services/frontend/package.json
```diff
@@ -74,5 +59,24 @@
...
+    "eslint-config-airbnb": "^18.2.1",
+    "eslint-config-prettier": "^7.1.0",
+    "eslint-plugin-import": "^2.22.1",
+    "eslint-plugin-jsx-a11y": "^6.4.1",
+    "eslint-plugin-prettier": "^3.3.1",
+    "eslint-plugin-react": "^7.22.0",
+    "eslint-plugin-react-hooks": "^4.2.0",
+    "lodash-webpack-plugin": "^0.11.6",
+    "prettier": "^2.2.1",
+    "webpack": "^5.12.1",
```
dynamic11, line 79: I don't see webpack as a dep before this change. why do we need it now?

mo-5: We needed it before but yarn does seem to display the errors. When I moved to npm it showed me the error. It's a requirement by the lodash bundle package that shrinks the size of lodash.

dynamic11 (commented):
I don't exactly understand why were switching back to npm vs yarn. can you please explainit?

mo-5:
Yeah so what I found is that with npm when you run with compatibility issues between packages it stops the installation. So if we have 2 packages share a package in between for example mom makes sure they share a package with the same version. However with yarn when you run into dependency tree related errors it still installs all the packages. For example if you go into our frontend right now and try npm install it will give you a dependency tree errors and that we need to downgrade some of our packages. For example we updated to react 17 but it's not yet supported by other packages we use.

dynamic11:
ok that makes sense. so are we going to migrate everything over to npm and drop yarn altogether? i dont want to create a mix of the two. Also on another note...a few months ago benoit and i had a discussion about the fact that we should be change our `yarn install` command since we want it to only install based on the lockfile and not the package.json for the purposes of consistency. Please see picture to see what I mean. please use this as inspiration to update this PR to do the same.

image

mo-5:
I could move it back to yarn since I already changed all the compatible versions together. Yeah in the end we will just use either yarn or npm we cant use the two. I don't mind either one. I could change it back to yarn and just keep the package versions the same.

dynamic11:
@MohamedRadwan I will leave the decision to you. you seem to understand this better than I do

MohamedRadwan:
This should be ready for review now. I moved back to yarn since it has parallel download for packages. I updated keycloak mock and made a release for it and added it to the package.json. all our current versions should be compatible.

MohamedRadwan:
This should be ready for review now. I moved back to yarn since it has parallel download for packages. I updated keycloak mock and made a release for it and added it to the package.json. all our current versions should be compatible.

mo-5:
This pr is good for review

PR #1212, opened 2021-01-08
[Feat] Searching ExFeeder in text will show ExFeeder nominees

Rizvida1:
#### Changes introduced

Searching ExFeeder in search bar will show exfeeder nominees

#### Related issue(s)

- Related to #1123


services/backend/src/core/search/util/getAllProfiles.js
```diff
@@ -398,6 +398,10 @@ async function getAllUsers(language, userId, request) {
         : undefined;
     }
 
+    if (info.exFeeder) {
+      info.exFeederText = "ExFeeder Ex-Feeder";
+    }
+
```
dynamic11: we also need the french. according to mena it has another term. please confirm with her

dynamic11 (changes requested):
please see my comment

services/backend/src/core/search/util/fuzzySearch.js
```diff
@@ -38,6 +38,7 @@ async function fuzzySearch(profiles, searchValue) {
       "teams",
       "telephone",
       "tenure",
+      "exFeederText",
```
dynamic11, line 41: you will also have to display this match in the fuzzy search match list on the result card

services/backend/package-lock.json
```diff
@@ -0,0 +1,14 @@
+{
```
dynamic11: why is this being added?

dynamic11:
can you also add a test to fuzzy search for this? because it is a corner case i want to make sure we have a test for it so we don't break it in t future.

dynamic11 (approved):

PR #1225, opened 2021-01-28
[Feat] Mandatory profile field for PRI number

### LINKED ISSUE #1219: Create a mandatory profile field for PRI number
The PRI number is a unique identifier used in another system called MyGCHR.  HRB would like to extract a copy and merge I-Talent Data with MyGCHR data that is stored on premise. 

We will need to add a mandatory field in the profile for employees to enter and update their PRI. 
 
Requirements:
Edit Profile
Must be mandatory for users who select to create profile
Exception: it cannot be mandatory for users that skip profile creation and go straight to search. However there needs to be a mandatory requirement the first time the user edits their profile.  
The field must only accept numbers and should not accept the same number (eg.99999999)
The visibility setting must permanently  set to private so that only the users sees it and the HR Admin role. 
Users can edit the PRI number in case they enter it incorrectly
Add "PRI" in the side nav bar on the edit page under the Contact Information Heading
Field and Side Nav must pass accessibility standard on the edit and profile pages (test with keyboard, tabing, and screenreader)

Profile Page
Display the PRI number above the "Employment Status" Section on the same row as the last edit date (left justify)
Use dark grey font (Eg: PRI:88888888 (A PRI has 8 numbers)

Profile Page
image

Edit Profile
image

image

Rizvida1:
#### Changes introduced

New PRI field on 

#### Related issue(s)

Closes #1219

#### Checklist
- [x] Update database diagram 
- [ ] Create new migration 

- [x] Update backend documentation 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible


mo-5:
can you just check why it's failing the tests?

mo-5 (approved):

mo-5:
@Rizvida1 this is my bad I didn't notice that you didn't create a migration for the database. In order to add the pri number to the database you will need to do the following: 

1. remove the changes you have made inside the migration. ( migrations are automatically generated and shouldn't be changed manually) 
2. add your changes to schema.prisma in src/database. 
3. run `yarn migrate:create` and give it a name

PR #1335, opened 2021-04-29
[New] Visibility Confirm Modal

### LINKED ISSUE #1125: Users forget to change visibility after creating or edit sections of profile
⚠️ **Based on user testing:**
- when users setup their profile they don't see/ forget that everything is set to private by default

ℹ️ **I propose**
- that on the last step ("finish step") we show the user s summary of everything that is visible or not with links to each profile section.

Rizvida1:
#### Changes introduced

Upon finishing a form, a modal will pop up showing the current visibility.

#### Related issue(s)

Closes #1125 

image
image

If the UI has been modified:
- [X] Has been tested on IE
- [X] It is accessible


MenaMachado:
@Rizvida1 , why does Employment Status say "connection"?

Rizvida1:
@MenaMachado "Connections" is one of the options of visibility. It can either be private, public or connections only.

MenaMachado:
@Rizvida1 , thanks for the reminder, I'm wondering if we should rename "connections" to "private group" for clarity but we can discuss that as a team and with the client. Also, must OL have 3 items?

Rizvida1:
@MenaMachado Thanks for pointing out the OL. It has been fixed.

MenaMachado:
@Rizvida1 - better!

services/frontend/src/components/profileForms/formControlButtons/FormControlButtonsView.jsx
```diff
@@ -19,7 +21,22 @@ const FormControlButtonsView = ({
   onReset,
   onFinish,
   fieldsChanged,
+  visibleCards,
 }) => {
+  console.log("ONFINISH", typeof onFinish);
```
mo-5: Do you need the console log here?

services/frontend/src/components/visibilityConfirmation/VisibilityConfirmationView.jsx
```diff
@@ -0,0 +1,112 @@
...
+    switch (cardStatus) {
+      case "PRIVATE":
+        return (
+          <>
+            <Tag color="red">
+              <FormattedMessage id="visibility.card.private" />
+            </Tag>
+          </>
+        );
+      case "PUBLIC":
```
mo-5, line 23: can this be changed to a function with variables being passed in for each case since its repeated code

mo-5 (changes requested):

mo-5 (approved):

dynamic11:
@Rizvida1 I just noticed that this error has been introduced due to this PR (I think):
image

Rizvida1:
> @Rizvida1 I just noticed that this error has been introduced due to this PR (I think):
> image

This is fixed in PR #1368

PR #1348, opened 2021-05-04
[Fix] update jenkins timeout and make package installation parallel 

mo-5:
#### Changes introduced

This should speed up the build processes a bit by making the package installation run in parallel instead of one by one. 

#### Checklist
- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:
- [ ] Has been tested on IE
- [ ] It is accessible

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,8 +6,8 @@ pipeline {
     }
 
     options {
+        timeout(time: 30) 
         disableConcurrentBuilds()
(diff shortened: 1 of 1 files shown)


mo-5:
Please note. In this branch I removed the testing and linting from the development branch build since we already run the testing and linting in the pull request. This would decrease the time we use Jenkins which would also decrease the time for a build to be deployed

dynamic11 (approved):

PR #1367, opened 2021-05-11
[Fix] create page for ie error using antd

### LINKED ISSUE #1366: redirecting IE to Edge not working on openshift
**Describe the bug**
When the app is hosted using openshift and opened in IE it automatically drops the file extension in the URL so the "ie-error.html" file does not open and the app gets stuck in the loop in the router.

Bug was introduced by #1318 

**To Reproduce**
1. Go to live website (on openshift) in IE
2. See the app redirect to /ie-error and not /ie-error.html

**Expected behavior**
It should show the message about the compatibility of IE in ie-error.html and redirect to Edge

dynamic11:
⭐  Changes introduced
Created an IE error page using antd to be more consistent with the rest of app. It removes the redirect to ie-error

ℹ️ Related issue(s)
closes #1366 

📷  Screenshots (if applicable)
image


mo-5 (approved):
LGTM

PR #1368, opened 2021-05-11
[Fix] Modal will close on error

### LINKED ISSUE #1361: Visibility setting is present even with errors
**Describe the bug**
I have noticed this will creatin a new profile that when you change something and save the visibility settings would be presented even with an error. These settings should only be presented once there are no errors.

**To Reproduce**

**Expected behavior**

**Screenshots**
image

**Additional context**

Rizvida1:
#### Changes introduced

The modal will close when an error appears.

#### Related issue(s)

Fixes #1361

### DIFF
--- a/services/frontend/src/components/visibilityConfirmation/VisibilityConfirmation.jsx
+++ b/services/frontend/src/components/visibilityConfirmation/VisibilityConfirmation.jsx
@@ -20,8 +20,12 @@ VisibilityConfirmation.propTypes = {
     PropTypes.oneOf(["PRIVATE", "CONNECTIONS", "PUBLIC"])
   ).isRequired,
   visible: PropTypes.bool.isRequired,
-  onOk: PropTypes.func.isRequired,
+  onOk: PropTypes.func,
   onCloseModal: PropTypes.func.isRequired,
+};
+
+VisibilityConfirmation.defaultProps = {
+  onOk: () => {},
 };
 
 export default VisibilityConfirmation;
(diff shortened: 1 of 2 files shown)


mo-5:
Should the modal even come up there is an error ? can you make it only come up if their profile is actually ready to be saved. Otherwise its not practical to present it since they will get it again after they fix the error. Do you know what I mean.

Rizvida1:
> Should the modal even come up there is an error ? can you make it only come up if their profile is actually ready to be saved. Otherwise its not practical to present it since they will get it again after they fix the error. Do you know what I mean.

The issue is that the form is validated upon submitting. So to do that, well have to create a separate form validation from the submission.

mo-5:
> > Should the modal even come up there is an error ? can you make it only come up if their profile is actually ready to be saved. Otherwise its not practical to present it since they will get it again after they fix the error. Do you know what I mean.
> 
> The issue is that the form is validated upon submitting. So to do that, well have to create a separate form validation from the submission.

Ah I see what you mean. I think we should leave it as is then since closing it immediately might cause more confusion. I will close the ticket. Sorry about that

Rizvida1:
> > > Should the modal even come up there is an error ? can you make it only come up if their profile is actually ready to be saved. Otherwise its not practical to present it since they will get it again after they fix the error. Do you know what I mean.
> > 
> > 
> > The issue is that the form is validated upon submitting. So to do that, well have to create a separate form validation from the submission.
> 
> Ah I see what you mean. I think we should leave it as is then since closing it immediately might cause more confusion. I will close the ticket. Sorry about that

I actually think this makes sense to leave in because it is more confusing if the modal stays open rather than return to the form. That way they can probably see the error.

mo-5 (approved):

PR #1473, opened 2021-06-16
[New] Turned Visibility Review modal to a drawer

### LINKED ISSUE #1443: visibility review modal not responsive in narrow screen
**Describe the bug**
please see screenshots

**Expected behavior**
it should be responsive

**Screenshots**
image
image

### LINKED ISSUE #1446: change format of table to review visibility settings and ensure it is more response (suggestion drawer from right)
When the table appears, it covers most of the profile in edit mode. Most of the settings are set at the sub-heading level and if someone is new to the app, it's not always obvious under which heading some of the sub headings belong (e.g job mobility). Given that the table doesn't have links to the sections (a consideration for future iterations), I think it would be better if the table followed a similar format as the nav bar. I think it would make it easier for users to find the section by quickly comparing the table to the side nav bar.  

Suggested format 2 columns instead of the current four; However the list would follow the same order as the nav bar.
image

Rizvida1:
#### ⭐ Changes introduced

Replaced Modal with more responsive and less intrusive drawer

#### 🔗 Related issue(s)

Closes #1446 
Fixes #1443 

image

#### ☑️ Checklist
If the UI has been modified:

- [X] It is accessible


dynamic11:
possible action item:
- save the changes to the profile before the confirm the drawer opens 
- change "save" to "confirm"

mo-5 (approved):
LGTM

PR #1495, opened 2021-06-22
[Fix] Official language save

### LINKED ISSUE #1470: Official languages save fails
**Describe the bug**
If you fill in the language form as in the screenshot and try to save you will see it fail to save with strange error

image

mo-5:
#### ⭐ Changes introduced

make status required

#### 🔗 Related issue(s)

closes #1470
#### ☑️ Checklist

- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:

- [ ] It is accessible


services/frontend/src/components/profileForms/langProficiencyForm/LangProficiencyFormView.jsx
```diff
@@ -293,6 +293,7 @@ const LangProficiencyFormView = ({
         <Form.Item
           name={statusName}
           label={<FormattedMessage id="lang.status" />}
+          rules={[Rules.required]}
```
dynamic11, line 297: can you add aria-required="true" to the input for accessibility. You can go to PrimaryInfoFormView.jsx for an example

dynamic11 (changes requested):
please see my comments about a11y

services/frontend/src/components/profileForms/langProficiencyForm/LangProficiencyFormView.jsx
mo-5: is aria-required only for input tags ? I was wondering because I cant find it used in Select tags https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute

dynamic11: I used it it on the primary info form and i tested with the screen reader and it is working. we just need the screen reader to pick up the "is required"

dynamic11 (approved):
looks good :)

PR #1513, opened 2021-06-28
[New] Column in the admin page to show profile visibility

### LINKED ISSUE #1463: Admin Dashboard, User Table - Create a separate colunm to identify hidden profiles
A user can set their profile to hidden. When they do, the status in the users table appears as hidden.  The status column has 2 options for an Admin: Active/Inactive. If an Admin sets the profile to Inactive, the profile no longer appears in search results. Only the user and Admin can see it.  There is no "hidden" option for an Admin. Therefore it can be confusing. 

It would be better to have a separate column  to identify which user profiles that have been hidden by the user. 

image

image

mo-5:
#### 🔗 Related issue(s)

closes #1463

image
image
image

#### ☑️ Checklist

- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:

- [ ] It is accessible

### DIFF
--- a/services/frontend/src/components/admin/userTable/UserTableView.jsx
+++ b/services/frontend/src/components/admin/userTable/UserTableView.jsx
@@ -354,13 +354,11 @@ const UserTableView = ({
       render: (record) =>
         record.status === "HIDDEN" || record.status === "INACTIVE" ? (
           <div className="icon">
-            <EyeInvisibleOutlined
-              style={{ fontSize: "150%", textAlign: "center" }}
-            />
+            <EyeInvisibleOutlined />
           </div>
         ) : (
           <div className="icon">
-            <EyeOutlined style={{ fontSize: "150%" }} />
+            <EyeOutlined />
           </div>
         ),
     },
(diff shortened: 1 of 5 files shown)


services/frontend/src/components/admin/userTable/UserTableView.jsx
```diff
@@ -320,14 +322,49 @@ const UserTableView = ({
...
-      render: (record) => renderStatusDropdown(record.key, record.status),
+      onFilter: (value, record) =>
+        value === "VISIBLE"
+          ? record.status === "ACTIVE"
+          : record.status === "HIDDEN" || record.status === "INACTIVE",
+      render: (record) =>
+        record.status === "HIDDEN" || record.status === "INACTIVE" ? (
+          <div className="icon">
+            <EyeInvisibleOutlined
+              style={{ fontSize: "150%", textAlign: "center" }}
```
dynamic11: can we move the styling to the less file?

dynamic11 (changes requested):

services/frontend/src/components/admin/userTable/UserTableView.less
```diff
@@ -12,3 +12,8 @@
   margin-left: 8px;
   padding-right: 10px;
 }
+
+.icon {
+  text-align: center;
+  font-size: 150%;
+}
```
dynamic11: be careful about using such a generic className since it will effect other parts of the project. 
i would either nest it inside another className or make a more specific name to avoid a conflict in styling

dynamic11 (changes requested):
please see my comment

services/frontend/src/components/admin/userTable/UserTableView.less
mo-5: Good point. Fixed

dynamic11 (approved):

PR #1516, opened 2021-06-29
New finish page

### LINKED ISSUE #1446: change format of table to review visibility settings and ensure it is more response (suggestion drawer from right)

Rizvida1:
#### ⭐ Changes introduced

Removed Visibility Review Modal.
Added Visibility Review to finish page and made finish page for both edit and create forms.

#### 🔗 Related issue(s)

Closes #1446

image

image

#### ☑️ Checklist

If the UI has been modified:

- [X] It is accessible

- [X] run `yarn i18n:validate" and clear errors


MenaMachado:
Hi Rizvi, 

If we're using "section" instead of link, I think the french may need to be updated to as follows: Cliquez sur n’importe quelle section pour la mettre à jour.

Also for that last sentence you can add a period in both languages.

Rizvida1:
> Hi Rizvi,
> 
> If we're using "section" instead of link, I think the french may need to be updated to as follows: Cliquez sur n’importe quelle section pour la mettre à jour.
> 
> Also for that last sentence you can add a period in both languages.

Hi @MenaMachado,

Is this what you meant?

image

image

Thanks.

dynamic11:
what do you guys think about this? I was just trying to create visual separation between all the text on the page
@MenaMachado @Rizvida1 
image
image

MenaMachado:
I like the change Ali proposed. It breaks up the content a bit. 

Can I propose a small edit.  

Can we say. "Below are your current visibility settings....." and in french you can say "Voici vos para....."

MenaMachado:
Rizvi, I just noticed in French it's n'importe "quelle" section pour la mettre à jour.  

Section is feminine and you need la mettre à jour.

Rizvida1:
Hi Mena,

This is what it looks like with said changes. 

image

image

The French has become a bit too large to fit within the area so shortening it might be a good idea.

Thanks

services/frontend/src/components/profileForms/doneSetup/DoneSetupView.jsx
```diff
@@ -1,52 +1,313 @@
...
+          </Descriptions>
+        ) : (
+          <Skeleton />
+        )}
+      </div>
       <Button
         icon={<SearchOutlined />}
         size="large"
-        style={{ marginRight: "25px" }}
+        className="mx-3 my-2"
```
mo-5, line 117: Can you use <Link> here instead of push

services/frontend/src/components/profileForms/doneSetup/DoneSetupView.jsx
```diff
@@ -57,6 +318,7 @@ const DoneSetupView = ({ userId }) => {
         icon={<UserOutlined />}
         size="large"
         type="primary"
+        className="mx-3 my-2"
         onClick={() => history.push(`/profile/${userId}`)}
```
mo-5, line 129: same here

mo-5 (approved):

services/frontend/src/components/profileForms/doneSetup/DoneSetupView.jsx
```diff
@@ -1,52 +1,313 @@
...
+                <Link
+                  to={editUrls.employmentEquityGroup}
+                  aria-label={` ${intl.formatMessage({
+                    id: "edit",
+                  })} ${intl.formatMessage({
+                    id: "employment.equity.groups",
+                  })}`}
+                >
+                  <EditOutlined className="mr-1" aria-hidden="true" />
+                  <FormattedMessage id="employment.equity.groups" />
```
mo-5: these are identical can you write a function and just call it for all edit urls that way we don't need to write each one separately

mo-5 (approved):

PR #1603, opened 2021-08-06
[Fix] More descriptive pri error message

### LINKED ISSUE #1519: [Perfective] PRI error message is not sufficient
### [Perfective] PRI error message can be revised
The PRI error message can be revised. It only says “Must be a valid PRI” but does not let the user know what is the required format. The telephone numbers on the same page do provide the user with the required format. PS the field which has the error are highlighted red and the error message is also red which is good

**Issue Location:** 
- Create Profile: Contact Information Form
- Edit Profile: Contact Information Form

**Steps to recreate:**
- Navigate to the Create Profile: Contact Information Form or  Edit Profile: Contact Information Form
- In the PRI text field type anything that is not an 8 digit number
- The error message will appear

EXP: The error message is descriptive and helps the user determine the expected input for the field
ACTUAL: The error message is missing the required format.

mo-5:
#### ⭐ Changes introduced

added an example for the pri error message

#### 🔗 Related issue(s)

closes #1519
#### ☑️ Checklist

- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:

- [ ] It is accessible


services/frontend/src/i18n/en_CA.json
```diff
@@ -272,7 +272,7 @@
   "rules.required": "Required",
   "rules.url": "Must be a URL address ex—https://www.google.com",
   "rules.valid.phone.number": "Must be a valid phone number",
-  "rules.valid.pri": "Must be a valid PRI",
+  "rules.valid.pri": "Must be a valid PRI ex—23241235",
```
dynamic11: I think we should change it to: Must be a valid PRI (e.g. 22222222)
* also i think we should give them a fake number that they can't actually use. in this case 22222222 is the correct number of digits but the validator wont accept it so the person can't use it to fake their PR)

services/frontend/src/i18n/fr_CA.json
```diff
@@ -272,7 +272,7 @@
   "rules.required": "Obligatoire",
   "rules.url": "Doit être une adresse URL ex—https://www.google.com",
   "rules.valid.phone.number": "Doit être un numéro de téléphone valide",
-  "rules.valid.pri": "Doit être un CIDP valide",
+  "rules.valid.pri": "Doit être un CIDP valide ex-23241235",
```
dynamic11: I think we should change it to: Doit être un CIDP valide (ex. 22222222)

dynamic11 (changes requested):

dynamic11 (approved):

PR #1666, opened 2021-09-23
[Fix] Remove /api/  from the api call and add it to the env variable 

mo-5:
#### ⭐ Changes introduced

to clean up the code a bit instead of having /api/ on every call we can just add /api in the env variables and call the api directly 
#### ☑️ Checklist

- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:

- [ ] It is accessible


services/frontend/src/i18n/en_CA.json
```diff
@@ -87,7 +87,7 @@
   "delete.diploma": "Delete all selected diplomas? They will also be deleted from user profiles associated with them.",
   "delete.school": "Delete all selected schools? They will also be deleted from user profiles associated with them.",
   "delete.skill": "Delete all selected skills? They will also be deleted from user profiles associated with them.",
-  "delete.user": "Delete selected user? This will also delete the user’s profile.",
+  "delete.user": "Delete selected user? This will also delete the user's profile.",
```
dynamic11: These translation changes are not related to the PR. Can we remove them?

dynamic11 (changes requested):
please see comments about translation file

dynamic11 (approved):

PR #1667, opened 2021-09-27
[Fix] Fixed console warnings from the education and experience forms

### LINKED ISSUE #1615: console error on education form
**Describe the bug**
please see screenshot

**To Reproduce**

1. Go to 'education form.'
2. open console 
3. See error

**Screenshots**
image

P4ul1029:
#### ⭐ Changes introduced

Changes address a console warning by antd that states after a form has been set by `initialValues`, those values can not be overwritten the `initialValue`  attribute from a child component.

#### 🔗 Related issue(s)

closes #1615 

No UI changes.


P4ul1029:
**A more in-depth explanation of the error and fix:**

The error warning is related to this: https://github.com/ant-design/ant-design/issues/24818

After retrieving a user's education data, the `intialValues` attribute for `QualificationsForm` is set. `EducationForm` is a child component of `QualificationsForm`. There is a `Form.Item` component inside `EducationForm` that has the `initialValue`  attribute that attempts to set `ongoingDate` to false. For newly added education entries that have not been saved, there is no error because no data is being overwritten. For previous education entries, antd logs a warning stating that `ongoingDate` can not be overwritten by `initialValue` because that value was already set by the `initialValues` attribute from `QualificationsForm`.

Removing the `initialValue` attribute from the `Form.Item` component inside `EducationForm` stops the warnings. However, for a new education entry that has not been save, `ongoingDate` is undefined and that causes a 422 error when attempting to save the entry. This means the `initialValue` attribute must be applied only to newly added education entries, and once they are saved, that attribute should no longer be applied because the initial value will be set by the `intialValues` attribute from `QualificationsForm`.

This was my original fix, but eslint doesn't allow props spreading, so the fix I had to implement is a bit repetitive:

services/frontend/src/components/profileForms/qualificationsForm/educationForm/EducationFormView.jsx
```diff
@@ -276,20 +276,36 @@ const EducationFormView = ({
                     )}
                   </Form.Item>
                   {/* Checkbox if event is on-going */}
-                  <Form.Item
-                    fieldKey={[fieldElement.fieldKey, "ongoingDate"]}
-                    initialValue={false}
```
dynamic11, line 281: You can use this I believe. It will help you avoid rapping the entire thing in the conditional statement and it will prevent repeated code:

 `initialValue={educationItem ? undefined : false}`

dynamic11 (changes requested):
please see my comments.

services/frontend/src/components/profileForms/qualificationsForm/educationForm/EducationFormView.jsx
P4ul1029: I tried it and I managed to reproduce the same warning:

However, the error didn't occur right when the page was loaded like before. It came after I tried creating and saving various education entries. I'm going to do some debugging to see if its the same problem as before or a new one.

dynamic11: I cant seem to recreate the error with the proposed changes. do you have step-by-step recreation instructions by any chance?

P4ul1029: After debugging, I discovered this error is sligtly different: when you delete an entry from the Education or Experience form and then add another one right away, the warning appears. It happens for both my code and yours, so I'll have to try and find a workaround to that.

P4ul1029:
The warning is no longer present when the page is initially loaded. However, there is an edge case where if you delete an entry from the Education or Experience tab and then add another one right after, the warning appears claiming you're overwriting the previous entry. The logging of this warning is more of an issue with antd than our code, with antd recommending to just ignore the warnings since they only appear on dev. Addressing the described edge case has been put on hold.

dynamic11 (approved):
LGTM

PR #1471, opened 2021-06-16
[Fix] GEDS endpoint can process missing fields

Rizvida1:
#### ⭐ Changes introduced

Added the ability to bypass missing fields received from GEDS API.

#### 🔗 Related issue(s)

Fixes #1469

PR #1727, opened 2021-11-19
[Fix] Remove Geds contact info check

mo-5:
#### ⭐ Changes introduced

I saw that Rizvi you have added this check as part of pr #1471 at first I thought it was to fix geds. However, our geds works by searching for the email so contact info cannot be null (if its null our geds will return an empty object). Please let me know if its needed. 

The reason I am removing it is because it creates a test branch which lowers our sonarqube score and its untestable.

please note: geds isn't currently working since all team members (I think they removed the contact information with the email change) 

#### ☑️ Checklist

- [ ] Update database diagram 
- [ ] Create new migration 

If the UI has been modified:

- [ ] It is accessible


dynamic11 (approved):
id we are sure this will not cause issues with empty phone numbers then im ok with it

mo-5:
If its an empty phone number then the value will be set to empty. What we are checking for is if the contact infromation is null which in our case would break geds since we use the email in contact infromation to get the profile

